PR #6, opened 2019-05-10
Update SDSS notebook for sdss_dr14 deployment

weaverba137:
This PR updates the SDSS notebook for the `sdss_dr14` deployment on the production server.

Tasks before merging:

- [ ] Final review and test.
- [x] Remove lines pointing to gp04.
- [x] Rename the `sdss_dr14` schema (one-line change).
- [x] Update HTML version of the notebook.

### DIFF
--- a/04_HowTos/FileService/How_to_use_SDSS_Data.html
+++ b/04_HowTos/FileService/How_to_use_SDSS_Data.html
@@ -11777,8 +11777,8 @@ div#notebook {
 <div class="inner_cell">
     <div class="input_area">
 <div class=" highlight hl-ipython3"><pre><span></span><span class="n">__author__</span> <span class="o">=</span> <span class="s1">&#39;Benjamin Alan Weaver &lt;baweaver@lbl.gov&gt;&#39;</span>
-<span class="n">__version__</span> <span class="o">=</span> <span class="s1">&#39;20190109&#39;</span>
-<span class="n">__datasets__</span> <span class="o">=</span> <span class="p">[</span><span class="s1">&#39;sdss_dr13&#39;</span><span class="p">,</span> <span class="s1">&#39;sdss_dr14&#39;</span><span class="p">]</span>
+<span class="n">__version__</span> <span class="o">=</span> <span class="s1">&#39;20190418&#39;</span>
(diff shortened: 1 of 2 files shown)


weaverba137:
See also noaodatalab/TapSchema#8.

weaverba137:
Last call for any review and test! @rnikutta ?

knutago:
The notebook How_to_use_SDSS_Data.ipynb completes without error.  Just a couple of comments:
1.  The first query took 1m54s to complete.  Perhaps it would normally complete much more quickly (none of the other queries took more than ~15 seconds), but this is close to the default timeout limit of 120 seconds.  Might be good to put an explicit timeout=300 in the query call to make sure that the query completes.
2.  In the cell investigating negative redshifts, I don't understand how the user can conclude "After looking at a few of these, you may notice that a number of them are point-like, red objects that are close to a brighter object. A plausible hypothesis is that CMASS color selection doesn't work so great when a nearby, brighter object might be contaminating the photometry, thus making it harder to distinguish LRGs from M stars.".  The links just point to a spectrum--where are you getting the object shape info and info on nearby objects?

The SDSS+Bokeh notebook may not be part of the work on this branch, but that notebook currently fails the import and setup cell.  JIRA Bug ticket filed.

weaverba137:
For item 2, did you actually click on the links?

knutago:
Yes, I clicked on a few, but not all.  That's why I made the comment.

weaverba137:
The point is that the links *don't* just point to a spectrum, there should also be an image.

weaverba137:
Ah, hang on, the image is on the skyserver side.  I will add instructions to follow the skyserver link.

knutago:
I see no images.  Screenshot:
Screen Shot 2019-05-15 at 11 30 05 AM

weaverba137:
I know.  The image is in the 'CAS' link.

knutago:
Ah.  Why not just link directly to that?

weaverba137:
I am working on it.

weaverba137:
OK, I've covered both the timeout issue and added direct links to SkyServer.  Those really are more effective at showing what's going on.

knutago:
Ok, this looks good.  Will merge and see about updating the /notebooks directory.  Should the branch be deleted?

weaverba137:
Thank you, deleting the branch.

PR #10, opened 2019-06-27
Update import calls

gnarayan:
`get_alert_ids` is now in `antares.dev_kit.fetch_data`


rnikutta:
Can you please update the version string of the notebook?
Can you please also write (here) what version of the antares module this notebooks works with? (It appears we may have that version deployed at DL already, since your updated import seems to work).
Thanks!

rnikutta:
Addendum: on DL machines:
```
pip3 list | grep -i antares
ANTARES                                    0.2.13
antares-client                             0.3.0
```
Can you confirm? (we also suggest that antares include a `__version__` module attribute. Thanks!)

PR #14, opened 2019-08-07
Notebooks 1 and 4-7 for LSST workshop

andyjarax13:
- Removed all .pyc files.
- Removed __pycache__ directories.
- Because of time constraints, I could not convert .png images to .jpg. I did however standardize .jpeg into .jpg. .png conversion will begin following the LSST event.
- I could not find the 'numpy as N' standardization error. I had already worked on standardizing imports away from the way they were originally written. Are there specific imports that should be changed?
- README file updated.
- Organized accessory data and image files into folders.
- Changed notebook filenames to better match their titles.


rnikutta:
In `05_Gravitational_Lensing/Data/lensdemo_funcs.py`. And in `05_Gravitational_Lensing/lensdemo_funcs.py`.
Which also brings up the question why the same code is there twice in two directories? Can you do away with one?
Finally: before I merge, you can push to the same branch (your master branch), and the additional changes will be included in the PR. I.e. no need to make a new PR each time :)

andyjarax13:
Thank you for pointing that out. I had never opened that file; I have only worked on the .ipynb files.
Also, thank you for the tip on the PR. I am still learning parts of git I have not had the chance to see yet!

PR #15, opened 2019-08-09
Update AntaresFilterDevKit.ipynb documentation

cstubens:


rnikutta:
Updated NB deployed on DL NB server.

PR #18, opened 2019-10-18
Improved ANTARES DevKit

gnarayan:
- new filter examples
- more documentation
- uses the latest version of DevKit API


gnarayan:
cc @cstubens @rnikutta

cstubens:
Thanks a bunch @gnarayan 👌

rnikutta:
@gnarayan Is it correct that you've also (heavily) updated a Teen Cafe notebook together with this? Thanks.

gnarayan:
Removed the teen astro cafe notebook - nothing was changed there, but the version had compiled output included. This might actually be a good idea for the EPO notebooks, since it's helpful to see the output with nbviewer without having to download and run, but that's for Connie and crew to decide.

PR #32, opened 2020-01-14
Add contributed science case notebook on TimeDomain sicence and Antar…

lchjoel1031:
…es example for variable stars (RCB stars).


rnikutta:
Thank you for contributing a NB to Data Lab!

Looks very good. Just a few suggestions to improve it further:

- Please follow our notebook naming scheme, i.e. CamelCase. Please also remove 'Demo' from the file name; rather, describe that the notebook is about, e.g. AntaresVariableRCBStars.ipynb

- Keep the section titles short. Have additional text as normal (markdown) text instead.

- Comments regarding science and the flow of the notebooks should be text in the md cells; comments in the code cells should just explain (very briefly) what the code is doing (if not clear); i.e., consider moving current code-comments into the md cells

So, e.g, the section title should be "Extracting light curve related properties", and the markdown subtext should be "Looks like we got a hit. Let's have a look at the last one (locus_id 425493). We first extract relevant properties (MJD, Mag, Mag_err) from the locus." Etc.

- It would be better if the auxiliary dat file was a standard csv file, i.e. have a header line, and have the columns be comma-separated, i.e.
```
ra,dec
109.663521,-24.948238
107.703934,-06.366807
273.270447,-25.029964
```
Also, remove the trailing white space from all lines.
Such a csv file can then be read with all standard csv-reading tools, even for other purposes.

- There's a lot of code repetition in this block
```
    for i in range(len(result_set)):
        if result_set[i]["survey"] == 1 and len(result_set[i]["properties"])>0 and result_set[i]["properties"]["ztf_fid"] == '1' :
            gdate.append(float(result_set[i]["properties"]["ztf_jd"]))
            gmag.append(float(result_set[i]["properties"]["ztf_magpsf"]))
            gerr.append(float(result_set[i]["properties"]["ztf_sigmapsf"]))
        if result_set[i]["survey"] == 1 and len(result_set[i]["properties"]) > 0 and result_set[i]["properties"]["ztf_fid"] == '2' :
            rdate.append(float(result_set[i]["properties"]["ztf_jd"]))
            rmag.append(float(result_set[i]["properties"]["ztf_magpsf"]))
            rerr.append(float(result_set[i]["properties"]["ztf_sigmapsf"]))
    if len(gmag)>0 and max(gmag)-min(gmag) > 2.: 
        print("Got a hit on locus_id: ",result_set[i]["locus_id"])
    if len(rmag)>0 and max(rmag)-min(rmag) > 2.:
        print("Got a hit on locus_id: ",result_set[i]["locus_id"])
```
E.g. the two `if` blocks under the `for` loop differ only on the 3rd logical comparison; this can be significantly simplified I hope? All `.append` are the same in both blocks, only the band differs... A good idea would be to define a function that does all checks an ops on an id supplied as an argument, and in the for loop just feed the ids to the function...

- Also, is there no "slicing" way to extract an array of numbers from `results_set`? I.e., instead of 

```
gdate = []
for i in range(len(result_set)):
    if result_set[i]["survey"] == 1 and result_set[i]["properties"]["ztf_fid"] == '1' :
        gdate.append(float(result_set[i]["properties"]["ztf_jd"]))
# etc
```

do something like this instead (or similar, I'm not familiar with the data structure of Antares search result):

```
idx = (result_set[:]['survey']==1) & (result_set[:]['properties']['ztf_fid'] == '1')
gdate = result_set[idx]['properties']['ztf_jd']
gmag = result_set[idx]['properties']['ztf_magpsf']
# etc
```
- Finally, the result figure needs axis labels and a legend. Plus a conclusion sentence (what are we looking at).

Thank you!
Robert

lchjoel1031:
Thanks for reviewing the notebook and giving suggestions, which are taken into account accordingly and the revised notebook has been upload with pull request. 

The index slicing is not working, and our software engineer suggests a new way to extract the values from the result_set.

Please kindly let me know if there is any question!

Cheers,
Chien-Hsiu

lchjoel1031:
Hi,

Just want to let you know that I've made another example notebook demonstrating how to use ANTARES alert database for moving objects:
AntaresSolarSystemObjectLightCurveExploration.ipynb

I put it under Time-domain instead of SolarSystem because this is using ANTARES and the science case is looking into the time-series photometry. Please kindly let me know if there is any question/concern.

Cheers,
Chien-Hsiu

rnikutta:
Many thanks Chien-Hsiu! Two tiny questions:
- I see two ipynb files in TimeDomain/AntaresExample/ Are you contributing both?
- AntaresSolarSystemObjectLightCurveExploration.ipynb seems to embed 809LundiaPhasedLC.jpeg but that file is only 7.5KB in size, and appears to be an html file, not jpg. Could you please fix it?
Thank you!

lchjoel1031:
Hi Robert,

> I see two ipynb files in TimeDomain/AntaresExample/ Are you contributing both?

Yes. 

>AntaresSolarSystemObjectLightCurveExploration.ipynb seems to embed 809LundiaPhasedLC.jpeg but that file is only 7.5KB in size, and appears to be an html file, not jpg. Could you please fix it?

It's my bad. Just upload the correct version. Please kindly let me know if it is still not working.

Cheers,
Chien-Hsiu

rnikutta:
Thank you very much for contributing the notebooks to Data Lab! Merged.

PR #63, opened 2020-09-02
Reuploaded: updates for Visualize SDSS Spectra w Bokeh

jacquesalice:


weaverba137:
@rnikutta, if you could please recover the file `sdss_viz_bokeh.py`, I think we ought to delay this PR until that file is present.

rnikutta:
@weaverba137 and @jacquesalice I have the file! Can you pull it or copy by hand from rnikutta branch?
https://github.com/noaodatalab/notebooks-latest/blob/rnikutta/04_HowTos/FileService/sdss_viz_bokeh.py
My recollection is that it does not yet work with the notebook b/c of the global variables it relies on.

weaverba137:
I've added the module.  I'll see if I can debug it quickly.

rnikutta:
Many thanks!

weaverba137:
Note that this branch seems to have been created from somewhere other than the master branch, because the master branch imports `sdss_viz_bokeh`, but this notebook does not.  I'm basically going to have to gut-replace the notebook on this branch, as well as debug the module.

rnikutta:
At some point I needed to "bless" the master branch to be able to release to auto-NB testing. I suppose that's where I forgot to git add the py file, apologies. It should be only required to import `sdss_viz_bokeh`, and of course fix how the NB interacts with it.

weaverba137:
This is now ready for testing.

rnikutta:
Thanks for this one @jacquesalice and @weaverba137 !
Runs fine except two things:

1) `ac.whoAmI()` needs to be commented out, and have the standard "If you need to log in to Data Lab ..." before the cell. And `ac` needs to be actually imported in the import cell.

2) The last cell ("Navigating through targets") just spins forever, and nothing happens
```
import time
sp.ispec = 0
for i in range(sp.nspec):
    sp.next()
    time.sleep(0.5)
```

Also, if this is to be an interactive cell, we need to comment it out, and add a sentence about uncommenting, so that the NB can run in auto-testing mode.

weaverba137:
Regarding #2: the code should be working fine, but it is not  obvious. You have to run that cell then scroll back up to the interactive plot to see it change. The documentation should describe that better.

rnikutta:
I propose then that we have the "viewport" (can be a new one) immediately above this cell, so that it's immediately obvious. OK?

weaverba137:
@rnikutta , it appears that you let #64 pass without making any specific requirements on having `ac.whoAmI()` commented out.  Therefore this notebook just copies what was done in #64.  I'll change it as requested.  You may want to think about a way to automatically check the boilerplate code in these notebooks, beyond just making sure the notebooks don't fail.

rnikutta:
Sorry I misspoke. ac.whoAmI() is fine, just needs to imported (it wasn't). When I wrote this I had ac.login() in mind. Apologies.

weaverba137:
OK, new version to test.

rnikutta:
Thanks, looks good. Only thing: the loop over spectra takes 2.9s per spectrum (plotting is slow I guess), and `sp.nspec` is 1000. So that's 48 minutes, way too long for the auto-testing. Would it suffice to loop instead over the first 20 spectra or so, for demonstration purposes? Thanks.

weaverba137:
Loop shortened.

rnikutta:
Perfect, thank you. Merging.

PR #73, opened 2020-11-06
Colab style updates

DavidVargasMora:
Hi @rnikutta!

These are some small font style and general structure changes that we were asked to implement. We would be glad if you could please check and merge in case everything is O.K.

Thanks a lot for all your help!


rnikutta:
Merged. Thank you!

DavidVargasMora:
Thank you so much, Robert!

PR #76, opened 2020-12-09
Add 04_HowTos/CrossmatchTables/How_to_use_pre_crossmatched_tables NB, gals.csv, and join.png

jacquesalice:


rnikutta:
Github won't show the diff for the ipynb file (too long), so my comments here:

- "The crossmatch tables at Astro Data Lab are structured as followed:" --> "are named as follows"
- Some DL URLs in the NB are noao.edu, others are noirlab.edu; noirlab.edu currently does not work, but we should have the correct SSL certificate any day now. Recommend changing all noao.edu URLs to noirlab.edu
-  In several places the return value from calling convert(response) is not captured in a Python object, but instead dumped directly into the NB.
 --> Recoimmend capturing the return value of convert() to an actual Python object, and then printing it (e.g. df = convert(...))
- In several places, replace 'ra' with 'RA', and 'dec' with "Dec" (when referring to 'right ascension' and 'declination'; not when referring to names of columns etc.)
- type (morphological model^), and below: ^PSF=stellar, REX=round exponential galaxy, DEV=deVauc, EXP=exponential, COMP=composite, DUP=Gaia source fit by different model.
--> Maybe on the same line?
- Where the SQL is written under "Write the query", a sentence or two about the double JOIN could be helpful.
- Under "Crossmatch a user-provided data table and a pre-crossmatched table", why is the query run once and the result shown straight in the NB, and then run again and saved to mydb? Better run the query only once (saving to mydb).
- The query under "Write a crossmatch query" (the 'qu' query) fails for me with:
```
"queryClientError: Error: syntax error at or near ";"
LINE 14:             ) as ss on true;
```
Weirdly, the very next run of the same query, but writing out to mydb, seems to work.
- Under "Speed test", the two queries, in two separate cells (good), would benefit from a clear reminder above each cell, which case this is (i.e. "First, running the cross-match ourselves", and "Now, the same but using pre-crossmatched tables"; or something like that.)
- Once all cells have been run, the NB's RAM footprint swells to over 4 GB. I think (and hope) that capturing the results of convert() to objects will improve the NB's garbage collection, and thus reduce the footprint. (alternatively, a `del df` here and there can take care of large and no longer needed objects)
- "Write query to randomly select 5 RA/dec targets" --> To make the sentence clearer, propose: "Write query to randomly select five targets (RA/Dec positions)"
- In the "Appendix", suggest to add a sentence of motivation, something like "A clear benefit of pre-crossmatched tables is that they contain the positions of the same objects in two datasets. We can use this to e.g. fetch images of an object from both surveys."
- I think `plt` is not imported.
- Appendix A3 needs a few words of motivation. What's happening, what goal, etc.
- The functions `make_cutout_comparison_table()`, and `_table2` and `_table3` surely look almost identical. Try to make a single function, with whatever changes between them simply being an argument to the function. After all, rule numero uno of good coding is "Thou shall not repeat thyself in thy code" ;-)) (Some suggestions: use `img1`, `img2` instead of `sd_img`, `de_img`; make `layer` an argument to the function; give default `None` values to the args that are optional (only needed in some calls). 
- Same with the `plot_cutouts()` function and its versions.
-

jacquesalice:
> * "The crossmatch tables at Astro Data Lab are structured as followed:" --> "are named as follows"

Changed.

> * Some DL URLs in the NB are noao.edu, others are noirlab.edu; noirlab.edu currently does not work, but we should have the correct SSL certificate any day now. Recommend changing all noao.edu URLs to noirlab.edu

Changed.

> * In several places the return value from calling convert(response) is not captured in a Python object, but instead dumped directly into the NB.
>   --> Recoimmend capturing the return value of convert() to an actual Python object, and then printing it (e.g. df = convert(...))

Good point. In these instances, they have been changed to: 
```
result = convert(response)
result
```

> * In several places, replace 'ra' with 'RA', and 'dec' with "Dec" (when referring to 'right ascension' and 'declination'; not when referring to names of columns etc.)

Changed.

> * type (morphological model^), and below: ^PSF=stellar, REX=round exponential galaxy, DEV=deVauc, EXP=exponential, COMP=composite, DUP=Gaia source fit by different model.
>   --> Maybe on the same line?

Changed to:
```
- type (morphological model: PSF=stellar, REX=round exponential galaxy, DEV=deVauc, EXP=exponential, COMP=composite, DUP=Gaia source fit by different model)
```

> * Where the SQL is written under "Write the query", a sentence or two about the double JOIN could be helpful.

Done.

> * Under "Crossmatch a user-provided data table and a pre-crossmatched table", why is the query run once and the result shown straight in the NB, and then run again and saved to mydb? Better run the query only once (saving to mydb).

Changed.

> * The query under "Write a crossmatch query" (the 'qu' query) fails for me with:
> 
> ```
> "queryClientError: Error: syntax error at or near ";"
> LINE 14:             ) as ss on true;
> ```
> 
> Weirdly, the very next run of the same query, but writing out to mydb, seems to work.

Just needed to remove the semicolon and it ran!
 
> * Under "Speed test", the two queries, in two separate cells (good), would benefit from a clear reminder above each cell, which case this is (i.e. "First, running the cross-match ourselves", and "Now, the same but using pre-crossmatched tables"; or something like that.)

Added.

> * Once all cells have been run, the NB's RAM footprint swells to over 4 GB. I think (and hope) that capturing the results of convert() to objects will improve the NB's garbage collection, and thus reduce the footprint. (alternatively, a `del df` here and there can take care of large and no longer needed objects)



> * "Write query to randomly select 5 RA/dec targets" --> To make the sentence clearer, propose: "Write query to randomly select five targets (RA/Dec positions)"

Changed.

> * In the "Appendix", suggest to add a sentence of motivation, something like "A clear benefit of pre-crossmatched tables is that they contain the positions of the same objects in two datasets. We can use this to e.g. fetch images of an object from both surveys."

Added.

> * I think `plt` is not imported.

Added to imports.

> * Appendix A3 needs a few words of motivation. What's happening, what goal, etc.

Added.

> * The functions `make_cutout_comparison_table()`, and `_table2` and `_table3` surely look almost identical. Try to make a single function, with whatever changes between them simply being an argument to the function. After all, rule numero uno of good coding is "Thou shall not repeat thyself in thy code" ;-)) (Some suggestions: use `img1`, `img2` instead of `sd_img`, `de_img`; make `layer` an argument to the function; give default `None` values to the args that are optional (only needed in some calls).

Very true! It has been reduced as such.

> * Same with the `plot_cutouts()` function and its versions.

Also reduced as such.

rnikutta:
Looks very good, many thanks!

Final three suggestions:

* The first time it's good to do the two-step thing:
```
result = qc.query(sql=query)
df = convert(result)
```
But in subsequent occurrences, maybe just use the built-in converter?
```
df = qc.query(sql=query,fmt='pandas')
``` 

* Then, regarding the speed test: I think it would be good to tailor the examples a bit with respect to how many rows are being cross-matched / joined, such that in the speed test both queries finish in reasonable time, but it's clear that one finishes in seconds, and the other takes minutes.

* In all the Appendix plots: the plotting of image thumbnails works also without converting the pandas entries with `.values`. I.e. instead of
```
list_ra1=r['ra1'].values       # ".values" convert to numpy array
list_dec1=r['dec1'].values
list_ra2=r['ra2'].values       
list_dec2=r['dec2'].values

img1,img2 = make_cutout_comparison_table(list_ra1,list_dec1,layer1,layer2,pixscale,list_ra2,list_dec2)
```
you can do straight:
```
img1,img2 = make_cutout_comparison_table(r['ra1'],r['dec1'],layer1,layer2,pixscale, r['ra2'],r['dec2'])
```
* And please disregard my earlier comment that the NB size was large. What I saw was the total RAM consumption of all my NBs open in JupyterLab! False alarm, sorry.

jacquesalice:
> * The first time it's good to do the two-step thing:
> 
> ```
> result = qc.query(sql=query)
> df = convert(result)
> ```
> 
> But in subsequent occurrences, maybe just use the built-in converter?
> 
> ```
> df = qc.query(sql=query,fmt='pandas')
> ```
Done.
> 
> * Then, regarding the speed test: I think it would be good to tailor the examples a bit with respect to how many rows are being cross-matched / joined, such that in the speed test both queries finish in reasonable time, but it's clear that one finishes in seconds, and the other takes minutes.

I didn't change this bc of the timecrunch to get out this notebook in time, but this is something I could change in the future. 

> * In all the Appendix plots: the plotting of image thumbnails works also without converting the pandas entries with `.values`. I.e. instead of
> 
> ```
> list_ra1=r['ra1'].values       # ".values" convert to numpy array
> list_dec1=r['dec1'].values
> list_ra2=r['ra2'].values       
> list_dec2=r['dec2'].values
> 
> img1,img2 = make_cutout_comparison_table(list_ra1,list_dec1,layer1,layer2,pixscale,list_ra2,list_dec2)
> ```
> 
> you can do straight:
> 
> ```
> img1,img2 = make_cutout_comparison_table(r['ra1'],r['dec1'],layer1,layer2,pixscale, r['ra2'],r['dec2'])
> ```
This is true, but in the plot_cutouts function I reference for those list_ra1[i], list_ra2[i] etc and when I tried to change it there it gave me an error for the appendix A3 cells... Again, I can look into this more at a later time but for now I will leave it as it is.
> 
> * And please disregard my earlier comment that the NB size was large. What I saw was the total RAM consumption of all my NBs open in JupyterLab! False alarm, sorry.

Ok good!

PR #77, opened 2020-12-15
Update broken async link

jacquesalice:


01_GettingStartedWithDataLab/01_JupyterPythonSQL101.ipynb
```diff
@@ -6,9 +6,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "__author__ = 'Robert Nikutta <nikutta@noao.edu>, Data Lab Team <datalab@noao.edu>'\n",
-    "__version__ = '20200821' # yyyymmdd\n",
-    "__datasets__ = ''\n",
+    "__author__ = 'Robert Nikutta <nikutta@noirlab.edu>, Data Lab Team <datalab@noirlab.edu>'\n",
```
rnikutta, line 9: Thanks for fixing the emails, but I don't think IT has enabled email aliases yet. Thus, nikutta@noirlab.edu doesn't work yet; it's robert.nikutta@noirlab.edu for now, alas. Could you please fix this where needed, for all addresses of course?

jacquesalice: Yes of course. Is that the same with everybody's? So Mike's would be michael.fitzpatrick@noirlab.edu and Adam's would be adam.scott@noirlab.edu? Will also have to change this in the User Manual edits...

rnikutta:
Thank you, merging.

PR #79, opened 2020-12-17
SpecClient HowTo/GettingStarted docs

mjfitzpatrick:
The HowTo and GettingStarted notebooks (and html versions) for the specClient interface.  Note these will not work as-is until the specClient is merged into the 'datalab' client and deployed to gp12.  Before that, get the 'specClient.py' from the 'specserver' repo and modify the "from dl import specClient as spec" line to just "import specClient as spec".


weaverba137:
I used this to get `specClient.py`:
```python
try:
    from dl import specClient as spec         # primary spectral data client interface
except ImportError:
    import os, sys
    sys.path.insert(0, os.path.join(os.environ['HOME'], 'specserver'))
    import specClient as spec

```
if you need it for your tests.

weaverba137 (changes requested):
Unfortunately, it's not practical to give line-by-line reviews of notebooks, so everything goes in this summary:

## Getting Started notebook

* NOAO Data Lab Team --> NOIRLab Data Lab Team, or Astro Data Lab Team
* In the section "What datasets are available?", the introductory paragraph has an unexpected change of font, like a code snippet wasn't closed properly. 
* `print(spec.catalogs(context='sdss_dr16'))`: In this list `sdss_dr16.segue1specobjall` is described as `SDSS DR14 spectra - Segue-1 survey`.
  - Also, minor quibble, but the official names are SEGUE-1, SEGUE-2, not Segue.

## How to use notebook

* Check the standard Data Lab boilerplate at the start of the notebook.  There are minor differences from the "Getting Started" notebook
* In the section "Dataset Contexts and Catalogs", the printouts of Python `dict()` objects are hard to read without whitespace.  Can those be printed differently so they are more informative?
* This notebook repeats some of the same material as the other notebook.  Maybe this is necessary, but there is a danger of these notebooks getting out of sync.  Can we consolidate the introductory material as much as possible and refer to the other notebook as needed?
* "All-Sky Queries and Mutltiple Catalog Fields"  --> "Multiple"
* In the second example under "Stacked Image Plot" one of the output strings is `'bytes' object has no attribute 'shape'`.  Is it possible to get rid of that?

rnikutta:
Just a quick add-on to one of Ben's comments: please use the NB template for general outline / layout of a notebook:

https://github.com/noaodatalab/notebooks-latest/blob/master/DataLabNotebookTemplate.ipynb

jacquesalice:
### Getting Started notebook
- Add the following to __datasets__ : ['sdss_dr16']
- Change the 'disclaimers1.php' link to 'disclaimers.php'
- I'm not sure why but the Summary, Disclaimer & attribution, Authentication, and Querying for spectra hyperlinks under Table of Contents didn't work for me
- Under Authentication, remove any mention of 'token'. The cell should just be: 
```
#authClient.login(input("Enter user name: (+ENTER) "),getpass("Enter password: (+ENTER) "))
authClient.whoAmI()
```
- Remove the following imports since they are not used in this notebook:
```
import pandas as pd
from IPython.display import Image
from dl.helpers.utils import convert 
``` 
- Under 'Querying for spectra', first paragraph last sentence misspelled identifers "ientifiers"

jacquesalice:
### How to use notebook
1. Under 'Summary', the API documentation link gives me a 404 error (since it doesn't exist yet?)
2. Change 'disclaimers1.php' to 'disclaimers.php'
3. Remove the following from imports since it is not used in this notebook:
`import pandas as pd`

rnikutta:
My comments provided via email (too lengthy for here).

mjfitzpatrick:
Without replying to each comment individually I'll say that each of the point has been addressed, and thank you for the comments.  The one thing not addressed are the comments that the HowTo is too long and that there is redundant information between the two notebooks.  I've added a 'References' section where each notebook points to the other, however I believe they each serve a different purpose and should be able to stand alone.  For example, 'contexts' are described in the HowTo and I think it would be confusing/annoying to have to lookup what they are in a GettingStarted doc in another part of the notebooks directory.  The only logical way to split up the (I don't consider too) lengthy HowTo would be to have separate Intro/Query/Access notebooks, but again I think these are distinct from the GettingStarted.

The repo has been updated with changes and the specClient.py change added to each of your notebook directories.  Server-side changes are deployed to gp06.

stephjuneau (changes requested):
Minor comments on "Getting Started with Spectra":

1- http://datalab.noao.edu/disclaimers.php --> noirlab

2- In the Summary, add that the science case notebook is called "EmLineGalaxies_SpectraStack.ipynb"

3- Under "Imports"
For saving query results in virtual [storage] <-- missing a word

4- Under "Querying for Spectra"
"an array of idkkjentifiers" <-- typo

5- Under "Preview Plots"
It may be confusing that displaying the png for the single spectrum preview has some cyan frame with numbers that seem to be pixel sizes. Not urgent but would check how to avoid it showing like this. Also would be nice to see some wavelength values and know whether the Preview can be called optionally in rest frame or in observed frame (depending on the use case, I can think of either one being desirable).

For the grid preview, would be nice to print in the notebook cell how many entries are in "id_list" so that the users can figure out how many pages there can go through. Also, can the user save the PNGs for future reference? If that's in the longer HowTo, then referring to that would be helpful.

mjfitzpatrick:
On Mon, Jan 4, 2021 at 5:09 PM Stephanie Juneau <notifications@github.com>
wrote:

> *@stephjuneau* requested changes on this pull request.
>
> Minor comments on "Getting Started with Spectra":
>
> 1- http://datalab.noao.edu/disclaimers.php --> noirlab
>
Changed.  ALso made 'https', similar changes to HowTo

2- In the Summary, add that the science case notebook is called
> "EmLineGalaxies_SpectraStack.ipynb"
>
Done


> 3- Under "Imports"
> For saving query results in virtual [storage] <-- missing a word
>
You may have an old notebook, this was previously fixed.


> 4- Under "Querying for Spectra"
> "an array of idkkjentifiers" <-- typo
>
Fixed


> 5- Under "Preview Plots"
> It may be confusing that displaying the png for the single spectrum
> preview has some cyan frame with numbers that seem to be pixel sizes. Not
> urgent but would check how to avoid it showing like this.
>
Fixed.  There was some additional styling in the HowTo that was removed.


> Also would be nice to see some wavelength values and know whether the
> Preview can be called optionally in rest frame or in observed frame
> (depending on the use case, I can think of either one being desirable).
>
I tried this before, the problem being the text was generally unreadable,
and if it was made large enough to read the file size became an issue.  The
fix may be do add some custom labelling or a plotPreview() method that uses
the pylot labels.  Either way would require regenerating all the previews
so I'd consider it a future feature enhancement. I created an issue so it
isn't forgotten.

> For the grid preview, would be nice to print in the notebook cell how many
> entries are in "id_list" so that the users can figure out how many pages
> there can go through. Also, can the user save the PNGs for future
> reference? If that's in the longer HowTo, then referring to that would be
> helpful.
>
I added a print of the number of images and an example of using the save()
method to save the PNG to a file to each notebook.  Also created a ticket
to do better annotation later.

Thanks for the comments!  I've updated the 'specserver' branch of the
'notebooks-latest' repo.

-Mike

PR #81, opened 2021-01-08
Add EmLineGalaxies_SpectraStack Notebook to Science Examples

Ragadeepika-Pucha:


rnikutta:
Hi @Ragadeepika-Pucha , thanks for the PR! It's a very nice NB. My comments below. You can just add new commits to this PR.

* Change the __version__ date string to current.
* "are hosted at the Data Lab" --> "are hosted at Astro Data Lab"
* "access the spectra of any given object" --> "spectrum" maybe (?) Or plural because it's more than one spectrum?
* Please change all `*noao*` URLs to `*noirlab*`; they should be working now.
* Under Imports, remove the text "# remove imports that are not needed"
* The `random` module is part of the Python std. library, i.e. please move this import into the "std lib" section of the imports cell.
* Can you please re-run the notebook, and then save, using the demo00 user? Otherwise 'raga_steph' shows up when the whoAmI() function is called.
* The sentence "H 𝛼 𝜆 6562, H 𝛽 𝜆 4861, [OIII] 𝜆 5006 and [NII] 𝜆 6583, to plot the traditional BPT diagram." seems incomplete?
* You can avoid having to escape the single quotes in `\'Infinity\')`, if you put the surrounding string in double quotes instead of single, i.e. `err_cond_sdss = ' AND \n'.join(["(E.Flux_{0}_Err > 0) AND (E.Flux_{0}_Err != 'Infinity')".format(e) for e in emission_lines])` (and several other places in the NB)
* When constructing a long (multi-line) SQL query string, you can avoid having to include explicit line break characters `\n` if for the full query string you use triple-double quotes (aka Python docstring quotes), e.g.
```
query = """
SELECT
E.ra as ra_ ...
"""
```
(and then the parentheses are not needed)
* "DataLab server" --> Data Lab server (with space)
* You don't need to change this now, can be done in a later update to the NB, but there are conventions in Python as to how to name functions and classes etc. For functions, it's very good that it describes an action (what it does), as e.g. your "def Get_SDSS_Spectra()", but the naming should be either all small caps with underscores, i.e. `def get_sdss_spectra()` or half-"CamelCase" (not sure what it's called colloquially, but like this `def getSDSSSpectra()`. I think most our NBs and code prefer the first version, and we use CamelCase for classes (e.g. `class UserName`).
* Very nice docstrings in your functions! Just one convention that should be followed: after the first triple-double quote, the single line should say what the function does. Then a blank line, then the rest. For example, in your first function, it should therefor be:
```
def get_SDSS_spectra(specobjid,  z, rest_frame = True):
    """Function to access the SDSS spectra of any object from the Data Lab file service.

        Returns the arrays of wavelength, flux, model and ivar.
        ...
    """
```
* Futher, in function docstrings, the input args should be under a "Parameters" section, and the output params under a "Returns" section. The reason is that for automatic dicumentation of functions with Sphinx, this is how it will be recognized correctly. Example:
```
def get_SDSS_spectra(specobjid,  z, rest_frame = True):
    """Function to access the SDSS spectra of any object from the Data Lab file service.

    Returns the arrays of wavelength, flux, model and ivar.
    Returns the rest-frame arrays if rest_frame = True.
    If rest_frame = False, returns the observed-frame arrays.
 
    Parameters 
    ----------
    specobjid : int64
       Unique SDSS spectrum ID.

    z : float
       Redshift of the source.

    rest_frame : bool
       Whether or not to return the rest-frame values. Default is True.
       
    Returns
    -------
    wavelength : array
       Wavelength array (Rest-frame values if rest_frame = True)

    flux : array
       Flux values array (Rest-frame values if rest_frame = True)

    [etc.]
    """
```
* Suggestion to make future life easier (e.g. with other NBs developed etc.): the lists of lines defined in function `Add_Lines()` as `emission_lines` and `absorption_lines` should be maybe read from a stand-alone text file? That way it's portable, reusable, and easy to update *once*, when an update is needed.
* A comment for later, about the data structure you chose to construct the lines lists... it seems unnecessarily complex, as a list of dictionaries. This leads to you then having to loop over them, lamba-mapping, just to get stuff out of the lists. A better suggestion could be:
  1) have the data comma-separated in a CSV file
  2) read in and populate for instance a single dictionary with the data, with dictionary keys only for name, lambda, emission, label.
But this is certainly for version 2.0 of your NB, and we can discuss with you and Stephanie down the road.
* When you first time plot a spectrum, your index is -500. This may be confusing to non-Python experts. Can you use a positive index value instead?
* Question for you and Stephanie: In the section "Function to align and stack spectra", in the function `Stack_Spectra()`, you align the spectra yourself. I though that Mike's spectro service has an option to returned aligned spectra already? Is that so?
* "The relative fluxes of [NII] 𝜆 6853 and and H 𝛼  lines is an indicator ..." --> Double "and", and "are and indicator"?
* "like an AGN, increases" --> no comma

PR #83, opened 2021-01-10
Initial DELVE DR1 access notebook

knutago:


kadrlica:
@knutago how are we doing with this? Waiting on the data to come out?

knutago:
> @knutago how are we doing with this? Waiting on the data to come out?

Yes as soon as table is live I will merge the PR.  Should be today!

PR #85, opened 2021-01-12
Stellar stream example notebook using DES DR1

kadrlica:
I plan to update this notebook for DES DR2 once the ingest is complete (should be as simple as changing the schema). Help with a review and merge would be appreciated @knutago @rnikutta (or anyone else). There may be some duplication with the dwarf galaxy examples.


rnikutta:
Thank you @kadrlica ! Merging. (After the AAS let's switch to DR2, and to the newer DL NB template (it's in the root dir).)

kadrlica:
@knutago suggests that we also move the notebook to `03_ScienceExamples/GalacticStructure`

PR #99, opened 2021-06-27
notebook updates

mjfitzpatrick:
Multi-spectra test id lists for DLS-13


rnikutta:
Tested from branch `fitz` in gp12, runs correctly, thank you. Merging.

PR #103, opened 2021-09-08
Reduce total runtime, remove errors

jacquesalice:
Previously the total runtime for this notebook was ~25 minutes and contained 2 timeout errors (on purpose). The new notebook now takes ~5 minutes to run and contains 0 timeout errors (replaced by a try/except statement).


rnikutta:
Thank you @jacquesalice !
Great runtime improvements, but I have several suggestions and requests before it can be merged:

- To compare "apples to apples", I propose we replace the LIMIT statements with a 'WHERE a.random_id BETWEEN 10 AND 10.2" or similar, when df1 and df2 are produced. That way we get the exact same rows (not just same number of rows). I.e.:
```
query1 = """SELECT
           a.source_id as id1,a.ra as ra1,a.dec as dec1,a.pmdec,a.pmra,a.w1mpro,a.w2mpro,
           gg.specobjid as id2,gg.ra as ra2,gg.dec as dec2,gg.z,gg.class,gg.veldisp,gg.veldisperr,
           (q3c_dist(a.ra,a.dec,gg.ra,gg.dec)*3600.0) as dist_arcsec 
         FROM 
            allwise.source AS a
         INNER JOIN LATERAL (
            SELECT s.specobjid,s.ra,s.dec,s.z,s.class,s.veldisp,s.veldisperr
            FROM 
                sdss_dr16.specobj AS s
            WHERE
                q3c_join(a.ra, a.dec, s.ra, s.dec, 1.5/3600.0)
            ORDER BY
                random()
            ASC LIMIT 1
        ) as gg ON true
        WHERE a.random_id BETWEEN 10 AND 10.25
```
and
```
query2 = """SELECT 
           X.id1,X.id2,X.ra1,X.dec1,X.ra2,X.dec2,X.distance as dist_arcsec,
           a.pmdec,a.pmra,a.w1mpro,a.w2mpro,
           s.z,s.class,s.veldisp,s.veldisperr
         FROM 
             allwise.x1p5__source__sdss_dr16__specobj as X 
         JOIN 
             allwise.source as a ON X.id1 = a.source_id 
         JOIN 
             sdss_dr16.specobj AS s ON X.id2 = s.specobjid
         WHERE a.random_id BETWEEN 10 AND 10.25
```
The resulting data frame rows are not in the same order, but this can be fixed later, e.g. via:
```
df1 = df1.sort_values('id1')
df2 = df2.sort_values('id1')
```
For the examples above with random_id between 10 and 10.25, I'm getting runtimes around 22 seconds (with warm cache, otherwise 1m40s), and 4 seconds, respectively.

- When a try/except block is used, an exception should be actually raised. I.e. not just a print() statement. E.g.:
```
try:
    df1 = qc.query(sql=query1,timeout=180,fmt='pandas')
except:
    raise
#    print("Query timeout at 180 seconds (3 minutes)")
```
Otherwise, a failed query can just look as if it returned in a second or two.

- In the two examples query1 and query2, the returned dist_arcsec values differ by 3600 (degrees to arcsec conversion), see attached screenshot.

dist

- Final point: I think that the double-JOINs and LATERAL JOINS are too complex to showcase the difference between crossmatching and using the pre-crossmatched tables. Why not just a simple two-table crossmatch?
And while at it, why not limit ourselves to just a few easy columns? (i.e. let's not confuse the NB reader with unnecessary complications, and let's focus on showing one thing, but very clearly).

Many thanks again!

rnikutta:
Hi @jacquesalice If the distance columns in the xmatch tables are fixed, can we pick up this NB again where we left off? (to implement the changes we discussed earlier, such that NB runtime is significantly reduced). Many thanks.

jacquesalice:
Yes, thank you for the reminder! I'll work on this today.

jacquesalice:
@rnikutta This is now ready for review.
Things changed:
- The Speed Test section now compares crossmatching on the fly and retrieving both tables' id, ra, and dec columns VS using a precrossmatched table to retrieve id1, id2, ra1, ra2, dec1, dec2, distance. Since the precrossmatched tables do not have random_id columns, I decided to use LIMIT 1000 for these queries. 
- I removed the try/except blocks since out of the 4 queries under Speed Test the longest one should run for a maximum of ~1min 10s. Let me know if you experience any errors. 
- The distance column issue has been fixed, see screenshot:
xmatch_dist
- Replaced ls_dr8 with ls_dr9
- The total runtime for this notebook is now ~2min.

rnikutta:
Hi @jacquesalice , thank a lot for the latest updates to the PR, looks very good! Could you please just make the following minor improvements? Afterwards, it's good to merge.

__In "How to use..." NB:__

- The language in the intro is probably verbatim from our Confluence page that describes what the pre-crossmatched tables should be. For this outward-facing NB I suggest we change the tone in a few places.
  - E.g., instead of "All crossmatch tables shall be minimalist, i.e. have only these columns: id1,ra1,dec1,id2,ra2,dec2,distance. Column descriptions in the crossmatch table shall contain the original column names in parentheses (makes it parsable)." I propose "All pre-crossmatched tables are only have these columns: id1,ra1,dec1,id2,ra2,dec2,distance. Column descriptions in the pre-crossmatched table contain the original column names in parentheses (this makes them parsable)."
  - "The column data types in a crossmatch table for columns id1 and id2 shall be retained from the mother tables." --> "The column data types in a pre-crossmatched table for columns id1 and id2 are retained from the mother tables."
  - "The data types for columns ra1, dec1, ra2, dec2 are DOUBLE."

- Change "The list of available crossmatch tables can be viewed on our query interface here under their respective schema." to "The list of available crossmatch tables can be viewed in our table schema browser under their respective schema." (remove here, and make "table schema browser" the link to https://datalab.noirlab.edu/query.php )
- Before the second comparison example, the section title is unclear: "For completeness, we switch the order of the queries and query from a different catalog." --> Completeness of what? Propose to say instead "Second example: we use a different catalog, and switch the order of queries"
- In both the first comparison example, you do compute the two DataFrames, with sorted rows in the end. But the DFs are not used for anything. You could for instance plot the two distance histograms to demonstrate that the results obtained by both the JOIN and the q3c crossmatch queries are identical. E.g.:
```
plt.hist(df1['distance'],histtype='step',color='b',lw=3,ls='-',alpha=0.5,label='Crossmatching manually')
plt.hist(df2['distance'],histtype='step',color='r',lw=4,ls=':',alpha=0.5,label='Using pre-crossmatched table')
plt.xlabel('angular distance (arcsec)')
plt.ylabel('number of sources')
plt.legend(loc='upper right',frameon=False);
```
- In the first appendix example with fetching image cutouts: are these random objects each time? I am getting different-looking images than last time...

__Examples NB__
- "from our 2 tables of interest" --> "from our two tables of interest"
- "we want to select from our 3 tables of interest" --> "we want to select from our three tables of interest"
- "RA/Dec of sdss object" --> "of SDSS object" (in multiple places in the NB)
- Format query strings to print better, for instance in this query string (but also in others):

```
query_single = ("""SELECT 
           X.ra1 AS ra_sdss, X.dec1 AS dec_sdss,
           S.z
         FROM sdss_dr16.x1p5__specobj__ls_dr9__tractor AS X 
         JOIN sdss_dr16.specobj AS S ON X.id1 = S.specobjid 
         WHERE X.ra1 BETWEEN %s and %s and X.dec1 BETWEEN %s and %s
         LIMIT 10000
         """ ...
```

to

```
query_single = ("""
SELECT 
X.ra1 AS ra_sdss, X.dec1 AS dec_sdss,
S.z
FROM sdss_dr16.x1p5__specobj__ls_dr9__tractor AS X 
JOIN sdss_dr16.specobj AS S ON X.id1 = S.specobjid 
WHERE X.ra1 BETWEEN %s and %s and X.dec1 BETWEEN %s and %s
LIMIT 10000
""" ...
```
- Under the MyDB example: "MyDB is a useful OS remote per-user relational database" --> remove "OS"

rnikutta:
Please ignore my comment "In the first appendix example with fetching image cutouts: are these random objects each time? I am getting different-looking images than last time...", the NB actually says that these are random sources.

PR #114, opened 2022-02-04
version from the afternoon before the Cafe

ameisner:
@rnikutta here's my updated version of the Colab notebook for tomorrow's Cafe. Thanks very much!


rnikutta:
Thank you for the PR.

PR #119, opened 2022-03-13
Adapted NB7 to work in Colab and simplified it 

DavidVargasMora:
Hi Robert,

I just uploaded some small modifications to adapt NB7 to work in Colab and simplified it according to Stephanie Juneau requests. Could you please check?

Thanks!


rnikutta:
Hi @DavidVargasMora thank you for the PR, merging.

PR #130, opened 2022-05-17
Update to SMASH dwarf notebook for NSF demo

knutago:
Updated SMASH dwarf notebook to include comparison query, updated SIA service to use general archive coadds, since SMASH-specific service is failing.  Needed for NSF demo.


rnikutta:
Merging, thanks!

PR #134, opened 2022-07-08
Teen Astronomy notebooks formatting edits

DavidVargasMora:
Hi @rnikutta,

We have been working with Mark Newhouse and Justine Schaen on several changes (mainly wording/formatting) for the Teen Astronomy Café notebooks. Could you please check and accept the pull request?

Thank you!


rnikutta:
Thank you @DavidVargasMora , merging. I noticed that all updates are to the e-TeenAstronomyCafe/ subfolder only. Is the "non-e" version of the NBs abandoned?

DavidVargasMora:
Thanks @rnikutta!
Yes, the "non-e" folder has not been touched since the beginning of the pandemic. 

We started working on the e-TeenAstronomyCafe when all things had to go virtual and decided to create a separate folder for the notebooks adapted to Colab.  I think we decided to preserve the original versions on the other folder, but not sure if we should keep it. What do you think @gnuhaus ?

PR #137, opened 2022-08-03
GOGREEN notebooks approved by Stephanie

felixpat10:
These are GOGREEN notebooks in .html and .ipynb files from 0 to 4 inclusive. There is an additional README.md file, making a total of 11 files.


rnikutta:
Thanks @felixpat10 and @stephjuneau , great job!

I have two questions / concerns, both regarding the last NB 4:

- When you plot the zoomed in spectrum of OII, and then again with the continuum subtracted, both plots look identical. Assuming nothing went wrong with plotting, it would seem that the original spectrum was already continuum-subtracted maybe?
- A bit further below, when you compute the difference between your estimations and those given by the GOGREEN team, the output prints:

```
Difference between calculated redshift and GOGREEN = 0.000388, percent error = 0.04%
Difference between calculated EW and GOGREEN = 51.091709, percent error = inf%.
Difference between calculated EW-model and GOGREEN = 0.761489, percent error = 1.51%
```
Did something go wrong in the second line? (see the inf result).

Please let me know if you want to fix these, and then amend this PR. Or if you want me to merge now, and you'll make another PR with any needed fixes. Thanks!

stephjuneau:
Hi Robert, this is showing a known bug in GOGREEN DR1 that will be fixed in DR2 and is explained in the text below that cell (e.g., when the ewoii is zero in the table, we point out that users can look at the ewoii_model column). I think we should leave it with the explanation and hopefully when we update to DR2 it will all be fixed as we anticipate!

stephjuneau:
Re your first question, the plot with continuum has a faint and nearly constant continuum but non-zero continuum. Unless we messed up the rendering order of the cell, it should show this.

rnikutta:
Thanks @stephjuneau

Indeed I missed the info about the inf result in comparing with GOGREEN released values, sorry!

On the issue of continuum subtraction... Ah yes, there is a constant offset on the y-axis. May I suggest that the plot function allow to take (ymin,ymax) and that both plots be scaled to the same range? Currently it does a variable `ax.set_ylim(-1e-18,max_spec+1e-18)`. It may also be useful to add to the plot function an `ax.axhline(0.)` to mark the zero level.

Please let me  know if I should merge now, or wait in case you want to make changes.

Thanks!

stephjuneau:
Hi @rnikutta   could you please merge when you get the chance? Felix will advertise the notebooks in his presentation and we will take care of the small change to notebook #4 in a separate PR (I also want to send along to Michael B. et al in case they might also suggest some small changes). Thanks!!

rnikutta:
@stephjuneau @felixpat10 Merged, thank you!

felixpat10:
Hi Robert,

There is a difference of about 0.6 Flux units, which can be seen where the blue line (ycontinuum) is shifted along with the actual spectrum. The first plot is at around 0.6 Flux, and the second is where the blue line and spectrum is at 0 Flux.

For your second question, we address that there is a bug within DL where the O II is not stored which results in a division of 0 when calculating the percent error. Therefore, we use the model O II, which is the third percent error.

If there is any confusion, please ask us.

Thank you,

Felix
________________________________
From: Robert Nikutta ***@***.***>
Sent: Wednesday, August 3, 2022 1:44 PM
To: astro-datalab/notebooks-latest ***@***.***>
Cc: Pat, Felix - (felixpat10) ***@***.***>; Mention ***@***.***>
Subject: [EXT]Re: [astro-datalab/notebooks-latest] GOGREEN notebooks approved by Stephanie (PR #137)


External Email

Thanks @felixpat10<https://github.com/felixpat10> and @stephjuneau<https://github.com/stephjuneau> , great job!

I have two questions / concerns, both regarding the last NB 4:

  *   When you plot the zoomed in spectrum of OII, and then again with the continuum subtracted, both plots look identical. Assuming nothing went wrong with plotting, it would seem that the original spectrum was already continuum-subtracted maybe?
  *   A bit further below, when you compute the difference between your estimations and those given by the GOGREEN team, the output prints:

Difference between calculated redshift and GOGREEN = 0.000388, percent error = 0.04%
Difference between calculated EW and GOGREEN = 51.091709, percent error = inf%.
Difference between calculated EW-model and GOGREEN = 0.761489, percent error = 1.51%


Did something go wrong in the second line? (see the inf result).

Please let me know if you want to fix these, and then amend this PR. Or if you want me to merge now, and you'll make another PR with any needed fixes. Thanks!

PR #138, opened 2022-08-10
Update antares notebooks to use the updated kernel

nesquivelr:
Hi 👋,

This PR updates the antares example notebooks along with the kernel they use. They no longer use ANTARES 0.4

Cheers,
Nicolás


rnikutta:
Sorry for the long delay @nesquivelr ! I was on vacation and missed this PR. Thanks!

PR #142, opened 2022-09-13
Update keywords file

jacquesalice:


rnikutta:
Thanks @jacquesalice, very nice! All the new keywords are actually used in the first cells of our NBs?

jacquesalice:
@rnikutta yep! They could be cleaned up a bit... For instance the use of both "sdss" and "SDSS", "spectra" and "spectroscopy", ... but I wanted to add them all to the list for now.

rnikutta:
Great, many thanks. Merging!

PR #145, opened 2022-10-07
Updated the emission-line notebooks to work with SPARCL

Ragadeepika-Pucha:
The notebook works well with the SPARCL. I still have to update the username to `demo00` before the final merging.


rnikutta (commented):
Thank you @Ragadeepika-Pucha !

Minor comments:

- "SPARCL" could be spelled out once. Probably best at the beginning of a NB.
- Import and instantiate like this:
```
from sparcl.client import SparclClient
client = SparclClient()
```
- Some section titles are all capitalized, some aren't. We try to standardize to not use capitalization everywhere. For instance: "Emission lines catalog from Portsmouth group". (same in the TOC)
- If the functions get_sdss_spectra() and 

In stack spectra NB:
- In section 'SDSS spectra': "Astro Data Lab" instead of "Astro Datalab"
- It really would be better if function plot_sdss_spectra() wasn't also fetching the spectra. Better to fetch outside, using get_sdss_spectra, then call the plot function to only plot.
- In section "selecting galaxies from the BPT" ("diagram" is missing btw), you define the rectangle borders twice: once for plotting and once for quering. Would be better to define them once and use for both the plotting and the querying. (reducing the chance of them going out of sync)

I also looked through the Outliers NB, looks good!

Thank you!

jacquesalice:
Hi @Ragadeepika-Pucha I looked through and ran both notebooks and they look awesome! I agree with Robert's comments and don't have any additional remarks.

Ragadeepika-Pucha:
Hi @rnikutta,

Thank you for the comments. I have updated the notebooks and modified the section names.

* For the plot_sdss_spectra() function in the stacking notebook, I wanted to have a function that directly plots the spectra given a specobjid. In the outliers notebook, I have it differently in the sense that the function takes in the spectra to plot. I wanted to show both the functions in the different notebooks. It is mainly because sometimes plotting is the main thing and one extra line can be avoided in the code.
* I am not sure I understand your rectangles comment. I define the rectangles only once (in cell 23) and select random sources within each rectangle only once as well (in cell 25). 

Let me know if you have any more questions/comments.
Thanks,
Raga

rnikutta:
Hi @Ragadeepika-Pucha , thanks for addressing my comments.
- Regarding plot function vs get_spectrum function: you avoid a line in calling a function... but now you have to define basically the same function twice... It's no problem, but it's not efficient either, hence my original suggestion.
- About defining the 4 rectangles/squares: you define the box edge values twice: in cell 22 in the `Rectangle()` instantiations. And then again in cell 23 (same numbers). If you now decided you want to change the exact boxes, you'd have to change the numbers in two places. Wouldn't it be better to define them only once, and then both create the `Rectangle()` instances in cell 22, and getting the index arrays in cell 23, using the only-once defined limits?
Thanks!

rnikutta:
Hi @Ragadeepika-Pucha and @stephjuneau Bumping the discussion on this PR; what do you think of my last comments from 12 days ago? Thanks!

stephjuneau:
Hi @rnikutta, Raga told me about these comments and that she would address them during one of the astro-coding sessions. Now that I read through, I'll take a look at the revisions as well so thanks for bumping them to make sure that they're still on the radar!

Ragadeepika-Pucha:
@rnikutta Thank you for explaining the comments, and I am sorry for the delay in submitting them.
I have separated the accessing and plotting the spectra functions and also updated the docstrings of the functions.
I have also added an extra cell to define the rectangles that are used for plotting and accessing the objects.
Let me know if any other changes need to be made. Thanks!

@stephjuneau please take a look at the updated notebooks. Thank you.

PR #146, opened 2022-11-01
Add retrieve_by_specid example

jacquesalice:


rnikutta:
Just a quick question before merging, @jacquesalice : why do we still have the (commented-out) cell that pip-installs sparclclient? At DL it is installed, and otherwise, for local use, the `from sparcl.client import SparclClient` indicates that the package must be installed.

jacquesalice:
@rnikutta fair point. We also plan to make a separate notebook on how users can use SPARCL without Data Lab, so it's not necessary to include in this notebook. I've removed the cell

rnikutta:
Thanks @jacquesalice , merging!

PR #148, opened 2023-01-11
Nb07 matplotlib update

DavidVargasMora:
Hi @rnikutta 

Colab updated the matplotlib version so I had to use an older release to fix the notebook. Could you please check and accept?

Thanks! Happy new year!


rnikutta:
Sorry @DavidVargasMora it took so long. I forgot about this PR (which happened when I was at a conference). Merging now!

PR #154, opened 2023-02-06
Temporary fix so matplotlib can plot

jacquesalice:
Since there is a bug in the GNIRS-DQS data files, Brandon Matthews has temporarily edited this NB so that any negative values are 0. This way, matplotlib can still plot the data


rnikutta (approved):
I confirmed that it runs and plots the scatterplot with errorbars, using Py3.8.8 and matplotlib 3.6.1. It still issues a warning about divide by zero though (see screenshot). Not critical, approving to merge.

gnirs-dqs-temp-fix

PR #155, opened 2023-03-01
Update to use Gaia DR3 instead of DR2

jacquesalice:


rnikutta (approved):
Looks good, thank you @jacquesalice !

PR #156, opened 2023-03-21
HVS WD science example notebook

jacquesalice:
This is a draft version. Still have to rephrase some descriptions and figure out why the second plot does not exactly match the Figure 3 plot in the paper


rnikutta (changes requested):
- Meta data in first cell: remove comments (they are for the template)

- Optional login cell:

  Change this:

    "If you need to log in to Data Lab, issue this command, and respond according to the instructions:"

    ```
    #ac.login(input("Enter user name: (+ENTER) "),getpass("Enter password: (+ENTER) "))
    ac.whoAmI()
   ```

  to this: (and maybe in the template, too):

    "If you need to log in to Data Lab, un-comment the cell below and execute it:"

   ```
    #ac.login(input("Enter user name: (+ENTER) "),getpass("Enter password: (+ENTER) "))
    #ac.whoAmI()
   ```

- Imports: sc, convert are not used in the NB. Remove their imports.

- Queries: check that all query strings that are to be submitted as ADQL actually used ADQL syntax (e.g. TOP instead of LIMIT).

- Query to select objects within 100pc: Why not select (in addition to some quality cuts) on: `... WHERE distance_gspphot < 100` (this field is given in parsecs)

- Section "Import the CSV files as Pandas Dataframes"

  I'd first say "We didn't run the actual queries, so we now read in the results from CSV files, using Pandas.

  Then in a second step, do the velocity selection.

- Transforming coordinates: using the trig functions from 'math' is a bit faster than numpy for small datasets, but the functions are not vectorized, so one has to loop (extra code).

  Instead, I propose to use numpy, and then the call doesn't need to loop. I.e.

```
  def galactocentric(parallax, l, b):
      "Calculate the Galactocentric distance and vertical distance."
      l = np.deg2rad(l)
      b = np.deg2rad(b)
      R_sun = 8.5
      r = 1/parallax
      X = R_sun - r*np.cos(l)*np.cos(b)
      Y = r*np.sin(l)*np.cos(b)
      Z = r*np.sin(b)
      R = np.sqrt(X**2 + Y**2)
      return R, Z
```

  And then computing R, Z is just:

  `R, Z = galactocentric(df_hvs['parallax'],df_hvs['l'],df_hvs['b'])`

  If you do that, then also 'import math' is not needed.

- In that section where X,Y,Z,R are computed, define these: "X,Y,Z are Cartesian coordinates, measured in parsec". "R,Z are cylindrical coordinates", etc. Say also what R_sun is, and what l and b stand for.

- Comments on motivating the steps of the science NB: some of the steps have either very short or no motivation. For instance:

  - Why are all these constraints applied to the queries?

   ```
    parallax_over_error > 4
    AND parallax > 0.25
    AND ruwe < 1.4
    AND ipd_frac_multi_peak <= 2
    AND ipd_gof_harmonic_amplitude < 0.1
    AND astrometric_sigma5d_max < 1.5
    AND phot_g_mean_mag - 5 * log10(1000.0 / parallax) + 5 > 6 + 5 * (bp_rp)
   ```

    Quality cuts? Selection criteria (motivated how?) Something else?

  - In Section "HR diagram for WDs and sdBs candidates", first paragraph:

  "Below is the Hertzsprung–Russell diagram for WD and sdB candidates that have the largest nominal two-dimensional velocities from Gaia DR3. High-velocity sdB candidates are classified as objects with 4≤𝑀𝐺≤6 and −0.5≤𝐺𝐵𝑃−𝐺𝑅𝑃≤−0.25."

  --> Why is that so? These selection criteria are mag and color-based, but supposedly select high-velocity objects? How so? Motivation / a bit of explanation is needed. (This is a science case NB, not just a "technique" NB).

  - The Section "Heights above the Galactic plane of the observed HVS WD candidates" also needs a bit more motivation. Why this plot? What do we expect to see? What doe we see?
  
    There's some of that in the text, but does not come across well-motivated. Things should follow from the previous steps.

rnikutta (approved):
Thank you @jacquesalice for the changes, and the extra context. Looks good, and ready to merge. Good work!

PR #160, opened 2023-04-13
Set half_sky=False in healpy.orthview() lest plotting fails.

rnikutta:
Hi @knutago , there's an unexplained regression with newer version of healpy (we had 1.16.2, and also downgraded to 1.16.1 but no cigar) that makes hp.graticule() calls fail, but the issue is more complex.

The tl;dr is that if in a previous call to hp.orthview() I change the argument half_sky=True to =False, the plot works (albeit showing both hemispheres, one empty). With -=True it fails.

The issue is complicated by other things. The observed behavior also depends on other parameters, e.g., having or not having the 'rot' arg set, and even the value of 'rot'.

This PR simply sets half_sky=False in the two NBs, to make them plot again. Would you mind signing off on this PR? I'd merge afterwards.

And if you can think of a better fix, or want to experiment a bit, please simply make another PR.

Thanks!


knutago (approved):
This seems like a sensible workaround.  Have you opened an issue with healpy?
Review is a little painful because the diff doesn't show, even though the change is very small.  I suppose because we store rendered notebooks in the repository.

rnikutta:
Thank you @knutago !

PR #167, opened 2023-04-27
Add new GMOS DRAGONS data reduction NB

rnikutta:
@jacquesalice could you please review this PR? I made only minimal changes to this latest version by Susan and Vini (e.g., update datestring, fixed a few typos, added "This will take 2 minutes" comments to two long-running cells, and cleared the outputs (b/c otherwise the NB swells to 7.5 MB instead of 200 KB or so without the outputs). Also added rendered HTML version.

### DIFF
--- a/04_HowTos/DataReduction/DragonsGMOS/DataReductionDragonsGMOS.html
+++ b/04_HowTos/DataReduction/DragonsGMOS/DataReductionDragonsGMOS.html
@@ -1,0 +1,15648 @@
+<!DOCTYPE html>
+<html>
+<head><meta charset="utf-8" />
+<meta name="viewport" content="width=device-width, initial-scale=1.0">
+
+<title>DataReductionDragonsGMOS</title><script src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.1.10/require.min.js"></script>
+
+
+
+
+<style type="text/css">
+    pre { line-height: 125%; }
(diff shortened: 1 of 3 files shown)


jacquesalice (approved):
Hi @rnikutta , this is great! I only saw 1 thing that needed to be changed (version date from '20240426' --> '20230427') so I uploaded a new version with that change to the PR.

I also updated keywords.txt to add "DRAGONS" and included it in this PR. 

I'm approving the PR, however, the last cell that is supposed to remove all of the FITS files and dirs didn't work for me. I even tried to change the directory in case it was looking in a different one:
```
import os
os.chdir('/home/jail/dlusers/ajacques/notebooks_20230421_204143/04_HowTos/DataReduction/DragonsGMOS/')
```
...but that still didn't remove them. Did it work for you?

jacquesalice:
Oh, sorry, I left the outputs in. I'll remove them and upload a new version

jacquesalice:
Done

rnikutta:
@jacquesalice Good catch; the cleanup cell doesn't work for me either. The reasons seems to be that the working directory is the root dir of the user, not the dir where the NB is. (but I'm somewhat surprised that the wget cell that fetches the data seems to be working fine. And it doesn't look like a permissions issue because in the terminal the files are owned by the user, and I can delete them there.

I'll email Vini and Susan, and Mike (maybe it's something with the DRAGONS kernel). Let's hold off on merging for the moment.

PR #170, opened 2023-05-24
New Extragalactic Tutorial (Milky Way Analogues)

simonmork:
Added a new Extragalactic tutorial detailing the process of finding Milky Way Analogues in the Legacy Survey.

@kadrlica


kadrlica:
@simonmork This notebook requires some minor updates before it would be ready to request a Data Lab review. In particular, I think that we might want to see if we can make a version that runs a bit more quickly. Let me know if you'd be interested in pursuing these updates over the summer.

kadrlica:
@simonmork I've made some updates to the notebook (including changing the name to something more descriptive). Take a look and let me know if this looks ok to you.

kadrlica:
@jacquesalice and @rnikutta: Both Simon and I have taken another pass through this notebook. Would you have a chance to start a formal review? Thanks!

rnikutta:
Hi @kadrlica and @simonmork , many thanks! We'll review ASAP (might be next week and the week after though, due to deadlines and travels).

jacquesalice:
Hi @simonmork and @kadrlica , thank you for the notebook! Here are my comments:
1. It looks like there are two notebooks in this PR: one called `analogues.ipynb` and one called `MilkyWayAnaloguesLSDR9.ipynb`. I assume the latter one is the one you intend on publishing? If so, make sure to remove the `analogues.ipynb` one from this PR.
2. Anywhere you have "Legacy Survey" change it to "Legacy Surveys".
3. In the Goals section when first mentioning the Legacy Surveys DR9, add a reference to their overview paper (Dey et al. 2019 https://ui.adsabs.harvard.edu/abs/2019AJ....157..168D/abstract)
4. In the Summary section, can you add a sentence or two about what Milky Way analogues are and why they're important/interesting?
5. Remove `from astropy.io import fits` and `import astropy.units as u` from the imports cell since they aren't used anywhere in the notebook
6. In the import cell, you have `import matplotlib as mpl` twice, remove one of these instances.
7. Also in the import cell, you have `from matplotlib import pyplot as plt` and then `import matplotlib.pyplot as plt`, remove one of these.
8. In the Data Access section, is there a reason why you split the query string into query1 and query2 to make query=query1+query2? Instead you could just do:
```
query = """
SELECT t.ra, t.dec, t.type, t.dered_mag_g, t.dered_mag_r, t.dered_mag_z,
       t.shape_e1, t.shape_e2, t.shape_r, p.z_phot_mean
FROM ls_dr9.tractor as t, ls_dr9.photo_z as p
WHERE t.ls_id = p.ls_id
      AND p.z_phot_mean > 0
      AND p.z_phot_mean < 0.1
      AND t.ra > {}
      AND t.ra < {}
      AND t.dec > {}
      AND t.dec < {}
"""
```
9. We offer pre-computed colors in the ls_dr9.tractor table so you don't have to compute them! Just add `t.g_r` and `t.r_z` to your query, then you can replace this:
```
g_r = g - r
r_z = r - z
```
with this:
```
g_r = data['g_r']
r_z = data['r_z']
```
10. In the Milky Way Analogue Selection section, you have `# MW size measurement from ???` , can you include where you got these values?
11. Also in the Milky Way Analogue Selection section when you plot the analogues on the color-magnitude diagram, the dots for the analogues are pretty hard to see.... May I suggest something like:
```
# Plot the selected MW analogues and MW values on the previous color-magnitude histogram.
fig=plt.figure(figsize=(6,6), dpi=120)
plt.hist2d(g_r_cut, M_r_cut, bins=[len(g_r_bins), len(M_r_bins)], cmap='jet', norm=LogNorm())
plt.errorbar(MW_g_r[0], MW_M_r[0], 
             xerr=sigma*np.array([[MW_g_r[2], MW_g_r[1]]]).T, 
             yerr=sigma*np.array([[MW_M_r[2], MW_M_r[1]]]).T, 
             c='r', marker='o', ls='', capthick=2, elinewidth=2, zorder=10,
             label="Milky Way")
plt.errorbar(MW_g_r[0], MW_M_r[0], 
             xerr=sigma*np.array([[MW_g_r[2], MW_g_r[1]]]).T, 
             yerr=sigma*np.array([[MW_M_r[2], MW_M_r[1]]]).T, 
             c='k', marker='o', ls='', capthick=4, elinewidth=4, zorder=4)
plt.scatter(g_r[milkyway_filter], M_r[milkyway_filter],
            c='white', marker='^', s=15, alpha=0.7, edgecolors='black',
            label='Milky Way Analogues')

plt.legend(loc='upper left')
plt.gca().set_xlabel(r'$g-r$')
plt.gca().set_ylabel(r'$M_r$')
plt.gca().invert_yaxis()
```
Which will look like:
Screen Shot 2023-07-13 at 12 44 59 PM

Overall, this notebook is very well structured, easy to follow, and is a really great example science case. Thank you so much for contributing it!

jacquesalice (changes requested):
Hi @simonmork and @kadrlica , thank you for the notebook! Here are my comments:
1. It looks like there are two notebooks in this PR: one called `analogues.ipynb` and one called `MilkyWayAnaloguesLSDR9.ipynb`. I assume the latter one is the one you intend on publishing? If so, make sure to remove the `analogues.ipynb` one from this PR.
2. Anywhere you have "Legacy Survey" change it to "Legacy Surveys".
3. In the Goals section when first mentioning the Legacy Surveys DR9, add a reference to their overview paper (Dey et al. 2019 https://ui.adsabs.harvard.edu/abs/2019AJ....157..168D/abstract)
4. In the Summary section, can you add a sentence or two about what Milky Way analogues are and why they're important/interesting?
5. Remove `from astropy.io import fits` and `import astropy.units as u` from the imports cell since they aren't used anywhere in the notebook
6. In the import cell, you have `import matplotlib as mpl` twice, remove one of these instances.
7. Also in the import cell, you have `from matplotlib import pyplot as plt` and then `import matplotlib.pyplot as plt`, remove one of these.
8. In the Data Access section, is there a reason why you split the query string into query1 and query2 to make query=query1+query2? Instead you could just do:
```
query = """
SELECT t.ra, t.dec, t.type, t.dered_mag_g, t.dered_mag_r, t.dered_mag_z,
       t.shape_e1, t.shape_e2, t.shape_r, p.z_phot_mean
FROM ls_dr9.tractor as t, ls_dr9.photo_z as p
WHERE t.ls_id = p.ls_id
      AND p.z_phot_mean > 0
      AND p.z_phot_mean < 0.1
      AND t.ra > {}
      AND t.ra < {}
      AND t.dec > {}
      AND t.dec < {}
"""
```
9. We offer pre-computed colors in the ls_dr9.tractor table so you don't have to compute them! Just add `t.g_r` and `t.r_z` to your query, then you can replace this:
```
g_r = g - r
r_z = r - z
```
with this:
```
g_r = data['g_r']
r_z = data['r_z']
```
10. In the Milky Way Analogue Selection section, you have `# MW size measurement from ???` , can you include where you got these values?
11. Also in the Milky Way Analogue Selection section when you plot the analogues on the color-magnitude diagram, the dots for the analogues are pretty hard to see.... May I suggest something like:
```
# Plot the selected MW analogues and MW values on the previous color-magnitude histogram.
fig=plt.figure(figsize=(6,6), dpi=120)
plt.hist2d(g_r_cut, M_r_cut, bins=[len(g_r_bins), len(M_r_bins)], cmap='jet', norm=LogNorm())
plt.errorbar(MW_g_r[0], MW_M_r[0], 
             xerr=sigma*np.array([[MW_g_r[2], MW_g_r[1]]]).T, 
             yerr=sigma*np.array([[MW_M_r[2], MW_M_r[1]]]).T, 
             c='r', marker='o', ls='', capthick=2, elinewidth=2, zorder=10,
             label="Milky Way")
plt.errorbar(MW_g_r[0], MW_M_r[0], 
             xerr=sigma*np.array([[MW_g_r[2], MW_g_r[1]]]).T, 
             yerr=sigma*np.array([[MW_M_r[2], MW_M_r[1]]]).T, 
             c='k', marker='o', ls='', capthick=4, elinewidth=4, zorder=4)
plt.scatter(g_r[milkyway_filter], M_r[milkyway_filter],
            c='white', marker='^', s=15, alpha=0.7, edgecolors='black',
            label='Milky Way Analogues')

plt.legend(loc='upper left')
plt.gca().set_xlabel(r'$g-r$')
plt.gca().set_ylabel(r'$M_r$')
plt.gca().invert_yaxis()
```
Which will look like:
Screen Shot 2023-07-13 at 12 44 59 PM

Overall, this notebook is very well structured, easy to follow, and is a really great example science case. Thank you so much for contributing it!

simonmork:
Hi @jacquesalice,

I have implemented the comments you mentioned in your above message. The only change I did not implement was #9 because the magnitudes that I pull I want to be dereddened and I am unsure if the precomputed colors in the tractor catalog are dereddened.

Let me know if there are any other final changes you want me to make!

-Simon

@kadrlica

kadrlica:
Thanks Alice and Simon. Sorry about the extra notebook, that was probably my fault. Do we want to commit all the cutout images?

rnikutta:
Thanks @simonmork . Indeed, the pre-computed colors appear to be using the not-dereddened mags, which can be confirmed like this:

```
df = qc.query("select mag_g, mag_r, dered_mag_g, dered_mag_r, g_r from ls_dr9.tractor limit 5",fmt='pandas')
df['my_g_r'] = df['mag_g'] - df['mag_r']
df['my_dered_g_r'] = df['dered_mag_g'] - df['dered_mag_r']
df

    mag_g       mag_r       dered_mag_g dered_mag_r g_r        my_g_r     my_dered_g_r
0   25.273912   24.131920   25.204136   24.084917   1.141993   1.141992   1.119219
1   25.448313   25.026280   25.378372   24.979166   0.422033   0.422033   0.399206
2   24.322979   23.566902   24.253012   23.519772   0.756077   0.756077   0.733240
3   26.226358   24.410812   26.156416   24.363699   1.815546   1.815546   1.792717
4   24.746489   24.035713   24.676622   23.988651   0.710775   0.710776   0.687971
```

I'll add my comments to the NB (if any) in the coming 2-3 days if that's ok.

rnikutta:
@kadrlica No, please don't commit any outputs other that the NB itself. Thanks.

simonmork:
@rnikutta I have recommitted changes deleting all exterraneous cutouts. Thank you for clearing up my confusion regarding the precomputed colors and reddening.

galaxyumi:
Hi @simonmork and @kadrlica, I am a new datalab team member. Thanks for your contribution! It is a very nice NB. 

Here are only minor suggestions:
1. In Goals, "SDSS-identified spiral galaxy Milky Way analogues" -> "SDSS-identified Milky Way analogues" or "SDSS-identified Milky Way-type spiral galaxies"
2. In Summary, "Milky-Way analogues are a unique class of object" -> "Milky Way analogues are a unique class of objects"
3. Can you fix the reference numbering to match flow? For example, the second reference appeared in Goals section is numbered as [4], not [2]. 
4. Missing unit on y-axis label in color-magnitude plots.

simonmork:
> Hi @simonmork and @kadrlica, I am a new datalab team member. Thanks for your contribution! It is a very nice NB.
> 
> Here are only minor suggestions:
> 
> 1. In Goals, "SDSS-identified spiral galaxy Milky Way analogues" -> "SDSS-identified Milky Way analogues" or "SDSS-identified Milky Way-type spiral galaxies"
> 2. In Summary, "Milky-Way analogues are a unique class of object" -> "Milky Way analogues are a unique class of objects"
> 3. Can you fix the reference numbering to match flow? For example, the second reference appeared in Goals section is numbered as [4], not [2].
> 4. Missing unit on y-axis label in color-magnitude plots.

Hi @galaxyumi, I have implemented the changes you listed in this message and have pushed them to my branch. Let me know if there is anything else you need.

rnikutta:
Hi @simonmork and @kadrlica ,

sorry for the stretched-out reviews... I finally found some time to take your NB fr a spin, it's very nice indeed! Thank you so much for contributing it to Data Lab. My comments below; they should be the last ones!

- Link any papers and external references directly. E.g., Licquia et al. 2016 [1]. You can either link directly (not need for footnotes), or link internally to the footnote, and from the footnote to the external resource.

- Replace anywhere in the text "LSDR9" with "LS DR9"

- In "Goals": "in order to locate Milky Way analogues to higher redshifts than many SDSS-identified Milky Way analogues." --> "in order to locate Milky Way analogues to higher redshifts than many SDSS-identified ones."

- In "Summary": "we query the datalab servers" --> "Data Lab servers"

- "sources that are best-fit with a galaxy profile" --> "best fit"

- In "Imports & Setup": I think you don't need `import warnings; warnings.filterwarnings("ignore")`, as I didn't see any warning without it. (Also, ignoring all types of warnings is not a good strategy, one might miss something unexpected yet important.)

- If you want the result of a query as, e.g., a Pandas dataframe, you can do this:

  ```
  q = "SELECT foo, bar FROM ..."
  response = qc.query(adql=q, fmt='pandas')
  ```

  I.e. you then also don't need to import the 'convert' helper function.

- Remove `%%time` from the one cell that uses it. Since recently we have a built-in cell execution timer installed.

- Nice solution you found to split the RA and Dec ranges into parts, run your query for each quadrant independently, and combine the results in the end. I must admit I was surprised that this was faster than running the query once, but on the entire RA and Dec range. Our DB specialist, and myself, have tried for some hours to find a faster way, but any solution required non-standard trickery, so we can't recommend it.

- For your looping over the RA and DEC slices, I suggest this simplification (simpler than the current logic):

  - Removed `time.sleep(10)`, it only slows down the NB.
  - Changed to `JOIN ON` syntax (the `WHERE id1 = id2` syntax is equivalent, but deprecated)
  - Used `BETWEEN` syntax
  - Simplified loop stepping logic
  - Requesting pandas DataFrame as output
  - Raised timeout because when the server is busy a query could time out

  Code to replace the entire looping query cell:

  ```
  width = 20
  RAMIN, RAMAX = 90, 130
  DECMIN, DECMAX = 40, 80

  dfs = []
  query = """
  SELECT t.ra, t.dec, t.type, t.dered_mag_g, t.dered_mag_r, t.dered_mag_z,
         t.shape_e1, t.shape_e2, t.shape_r, p.z_phot_mean
  FROM ls_dr9.tractor as t
  JOIN ls_dr9.photo_z as p
  ON t.ls_id = p.ls_id
  AND p.z_phot_mean BETWEEN 0 AND 0.1
  AND t.ra BETWEEN {} AND {}
  AND t.dec BETWEEN {} AND {}"""

  for ra1 in range(RAMIN,RAMAX,width):
      for dec1 in range(DECMIN,DECMAX,width):
          t1 = time.time()
          q = query.format(ra1, ra1+width, dec1, dec1+width)
          response = qc.query(sql=q, fmt='pandas', timeout=600)
          dfs.append(response)
          print('FINISHED FOR RA %d-%d, DEC %d-%d AFTER %d s' % (ra1,ra1+width,dec1,dec1+width,time.time()-t1))

  df=pd.concat(dfs,axis=0)
  df
  ```

- Please adjust the description of the looping procedure (just under the headline "Data Access") accordingly.

- Please follow PEP8 recommendations regarding whitespace around operators, assignments and in variables tuples.
  https://peps.python.org/pep-0008/#whitespace-in-expressions-and-statements

  Examples:

  - This:
     
      ```masterfilter=np.ones(len(arrays[0]),dtype='bool')```

    should be:

      ```masterfilter = np.ones(len(arrays[0]),dtype='bool')```

  - This:

      ```ra,dec,width=i,j,int(SKIP/2)```

    should be:

      ```ra, dec, width = i, j, int(SKIP/2)```  

    etc.

    Note that there's no whitespace around '=' in keyword arguments to functions.

- If you're using a function only once, don't define it. Functions are for re-use. I.e., remove `def noni()` and have the code directly in the cell.

- Also, the mechanics of defining your filters are unnecessarily complicated. The entire cell that contains `def noni()` can be replaced with this:

  ```
  ninfilter = df['dered_mag_g'].notnull() & df['dered_mag_z'].notnull() & df['dered_mag_z'].notnull()
  typfilter = (df['type'] == 'DEV') | (df['type'] == 'EXP') | (df['type'] == 'SER')
  data = df[ninfilter & typfilter]
  print(len(data),'galaxies after quality cuts')
  ```

- Several comments on this cell:

  ```
  g = data['dered_mag_g']
  r = data['dered_mag_r']
  z = data['dered_mag_z']
  g_r = g - r
  r_z = r - z

  redshifts = np.asarray(data['z_phot_mean'])
  abs_e = (data['shape_e1']**2+data['shape_e2']**2)**0.5
  a = (1-abs_e)/(1+abs_e)

  half_light_rad = np.pi/180/60/60*data['shape_r']
  half_light_kpc = cosmo.angular_diameter_distance(z=redshifts).value*1000*half_light_rad

  cosmo_correction = cosmo.distmod(z=redshifts).value + 5*np.log10(cosmo.h)
  M_g = g - calc_kcor('g', data['z_phot_mean'], 'g - r', g_r) - cosmo_correction
  M_r = r - calc_kcor('r', data['z_phot_mean'], 'g - r', g_r) - cosmo_correction
  M_z = z - calc_kcor('z', data['z_phot_mean'], 'r - z', r_z) - cosmo_correction
  ```

  - The code lines computing 'a' need a bit more context. What's happening, and why? Either ad a line of explanation to the markdown cell above the code cell, or add a comment line just before the code:

  ```
  # Compute XYZ, which is the ABC (and add reference if possible)
  abs_e = (data['shape_e1']**2+data['shape_e2']**2)**0.5
  a = (1-abs_e)/(1+abs_e)
  ```

  - Same for the two lines computing the halflight radius.

  - Think if you could use astropy.units; and if not desired, at least state the units for each line

- At the end, before "References", add a "File Cleanup" cell.
  - Text below the Section headline. E.g.: "To remove the cutout generated images from the directory, un-comment the cell below and run it."
  - Followed by a commented-out code cell:

    ```
    # !rm -f `pwd`/*png
    ```

simonmork:
Hi @rnikutta,

Thank you for your detailed feedback. I have implemented your requested changes, and so the notebook should be ready for a final review and merge. Let me know if there is anything else I need to do. Thanks again!

rnikutta:
Thanks @simonmork ! A few items have not been addressed though:

- Link to any and all references (e.g. the Licquia et al. 2016 paper, the Dey et al. 2019 paper, etc. (specifically, in the Goals and Summary sections)
- I don't think your loop logic
```
RAMIN, RAMAX, RACHUNKS = 90, 120, 3
DECMIN, DECMAX, DECCHUNKS = 40, 70, 3

ra_width = (RAMAX-RAMIN)/RACHUNKS
dec_width = (DECMAX-DECMIN)/DECCHUNKS
for ra in np.linspace(RAMIN, RAMIN+ra_width*(RACHUNKS-1), RACHUNKS):
    for dec in np.linspace(DECMIN, DECMIN+dec_width*(DECCHUNKS-1), DECCHUNKS):
```

is simpler than the one I proposed:

```
width = 20
RAMIN, RAMAX = 90, 130
DECMIN, DECMAX = 40, 80

for ra1 in range(RAMIN,RAMAX,width):
    for dec1 in range(DECMIN,DECMAX,width):
```

Why do you prefer it?

- Replace this wording:
" Since we are limited by the amount of data that any one query is allowed to take at any given time"
with
" Since we are limited by the maximum runtime of a query"

- Replace "In order to use the selection cell for arbitrary RA and Dec, change the following values:" with "Change the following values to adjust for your use case:"
- Fix typo in "half-light radiys"

Everything else looks great!

simonmork:
Hi @rnikutta, sorry about the link mixup and typos, those should be up to snuff now.

As for the loop logic, I've left in place my most recent implementation for three reasons.

- The end user can query for rectangles of arbitrary aspect ratio, since the loop automatically computes the chunk width in both the RA and Dec directions, whereas the suggested loop is limited to square chunks. 
- It allows for decimal values of width, whereas the suggested loop with the python-native `range()` function only works for integers.
- Using the suggested loop is unstable if `(((RAMAX-RAMIN)%width == 0) and ((DECMAX-DECMIN)%width == 0)) == False` since it will overshoot the specified RAMAX and DECMAX in these instances. See the below image for an example.

Screenshot 2023-09-18 220846

I appreciate your detailed follow-up very much!

jacquesalice (approved):

rnikutta (approved):
Thank you @simonmork ! And thanks for explaining the loop motivation, make sense. Merging!

kadrlica:
Congratulations @simonmork! Thank you reviewers for all your detailed feedback!

PR #172, opened 2023-05-25
initial commit of clustering redshifts

rbgdet:


kadrlica:
@rbgdet This notebook requires some updates before it would be ready to request a Data Lab review. In particular, it needs to be follow the Data Lab formatting requirements, and it requires the addition of some explanatory text to explain the goals and approach that are being taken. Let me know if you'd like to continue with this process, and we can work toward an updated version over the summer.

kadrlica:
I don't think there are any plans to further develop this notebook, so I think this PR can be closed (@jacquesalice @rnikutta).

PR #180, opened 2023-06-13
revised SDSS EmLine SpectraStack

stephjuneau:
I made the fix (added a WARNING message when the returned number of records differs from the requested number), and updated both the ipynb and html version. I runs well with default Py3 kernel.


stephjuneau:
FYI @Ragadeepika-Pucha & @jacquesalice I have added a fix with a Warning when the number of SPARCL records differ from the requested number (but then the code proceeds with the returned records). We probably won't need to change the notebook because the change is a healthy sanity check to have, but we should run it again after the next SDSS/BOSS re-ingest in the future!

rnikutta (approved):
Tried on gp12, runs good! Thank you @stephjuneau . Merging.

PR #183, opened 2023-06-19
testnotebooks.py code now traps two kernel exceptions: DeadKernelErro…

rnikutta:
…r (likely due to RAM exhaustion), and NoSuchKernel. The test suite continues with the next NB in both cases, rather than quiting.

### DIFF
--- a/tests/testnotebooks.py
+++ b/tests/testnotebooks.py
@@ -325,6 +325,9 @@ def run(paths,include=('/**/*.ipynb',),exclude=('/**/*_tested.ipynb',),plain=Fal
         except NoSuchKernel:
             cprint('REQUIRED KERNEL NOT PRESENT','red')
             test = 'FAIL'
+        except Exception as e:
+            cprint('AN EXCEPTION OCCURRED: %e' % e,'red')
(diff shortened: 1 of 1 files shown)


iglesu (commented):
The only review I'd made is if we want to trap any generic exception, DeadKernelError and NoSuchKernel are the expected ones but what about the unexpected?
e.g. adding last 

except Exception as e:
    cprint(f"KERNEL DIED UNKWN EXC: {e}", 'red')

rnikutta:
Thank you @iglesu ! Added generic Exception trap. Please approve if you are satisfied.

iglesu (commented):
cprint('AN EXCEPTION OCCURRED: %e' % e,'red')

Isn't "%e" used for scientific notation?, I might be wrong but a %s should suffice
cprint('AN EXCEPTION OCCURRED: %s' % e, 'red')

rnikutta:
Ouch, a typo! Thanks for catching it @iglesu . Fix committed, please review/approve.

iglesu (approved):
Great! thank you!

PR #185, opened 2023-06-22
Update keywords

jacquesalice:
@rnikutta In this PR:

- Added `sparcl` to keywords.txt
- Added `sparcl` to list of keywords in the notebooks that use SPARCL
- Added links to science use-case NBs that use SPARCL in the SPARCL How-To NB


rnikutta (approved):
Thanks, looks good. Merging.

PR #190, opened 2023-08-04
DESI EDR Tutorial Notebook with Python 3 Kernel

Ragadeepika-Pucha:
@rnikutta @stephjuneau @jacquesalice @adambolton - This PR is for the DESI EDR notebook that will work with Python 3 Kernel. 

Thanks,
Raga


jacquesalice (changes requested):
Hi @Ragadeepika-Pucha thanks for this! Here is my feedback:

1. In the first cell that lists the authors, tags, etc remove the commented out parts so it's just:
```
__author__ = 'Ragadeepika Pucha <rpucha@arizona.edu>, Stephanie Juneau <stephanie.juneau@noirlab.edu>'
__version__ = '20230804' # yyyymmdd
__datasets__ = ['desi_edr']
__keywords__ = ['sparcl', 'spectra', 'database']
```
2. In the Table of Contents, change "References and Acknowledgements" --> "References and Acknowledgments"
3. In the Goals section, change "SPectral Analysis and Retrievable Catalog Lab" --> "SPectra Analysis and Retrievable Catalog Lab"
4. In the Disclaimer and Attribution section, change "https://datalab.noao.edu/disclaimers.php" --> "https://datalab.noirlab.edu/disclaimers.php"
5. Anywhere you have "DataLab" change it to "Data Lab"
6. In the DESI-EDR Database section, you mention "We will focus on three different tables in this notebook:" but then go on to show only the zpix and target tables. Does this description need to be updated to how it is in the other notebook that uses the DESI kernel? I.e. "We will focus on the zpix table in this notebook. This is the main redshift catalog from the DESI survey....."
7. In the "Selecting Sources based on Targeting Information" section, I think you can remove this part, since you're only getting the targeting information from the zpix_cat table in this notebook:
```
# sv1_targetmask.desi_mask corresponds to SV1_DESI_TARGET
# sv2_targetmask.desi_mask corresponds to SV2_DESI_TARGET
# sv3_targetmask.desi_mask corresponds to SV3_DESI_TARGET
```
8. Also in the "Selecting Sources based on Targeting Information" section, change "targetting" --> "targeting"
9. Also in the "Selecting Sources based on Targeting Information" section, when you list the six main classes on DESI targets, add a bullet point for "SCND: secondary targets" and make it six main classes. 
10. In the "Other masks" section, `desi_mask` isn't used in this notebook so that sentence should be changed to something like "The `sv*_desi_target` columns consist of...."

stephjuneau (changes requested):
Hi @Ragadeepika-Pucha,

I agree with Alice's comments. In addition, here are a few more minor comments:

### Everywhere

Replace "DESI-EDR" with "DESI EDR"

### Goals

"but does not require any DESI-related software. It can be run using only the Data Lab and SPARCL dependencies." --> "but it has been modified such that installing the DESI software (desispec) is not necessary to run this notebook. As a result it can be run using a Python 3 kernel with only the Data Lab and SPARCL dependencies."

### Summary

"Kitt Peak, AV" --> "Kitt Peak, Arizona"

"This includes SV1, SV2, and SV3 (the One-Percent survey), which is hosted by the Data Lab with the desi_edr database. " --> "The latter include three phases of survey validation: SV1, SV2, and SV3 (the One-Percent survey). The DESI EDR catalogs are loaded as tables in the desi_edr database. "

"More details ... [are] available here."

### DESI EDR Database

You might not need cell [7] that shows the target table since it is not being used.

### References

"the Myers et al. 2023" -->  "DESI target selection pipeline paper (Myers et al. 2023)"

Ragadeepika-Pucha:
@stephjuneau @jacquesalice @rnikutta @adambolton - updated both Py3 notebook with the above suggestions. Fixed some small things in the DESI kernel notebook.

stephjuneau (approved):

stephjuneau:
Thanks @Ragadeepika-Pucha! In my opinion, this is now ready to merge (by @jacquesalice or @rnikutta ?)

rnikutta:
Taking a look right now...

rnikutta (changes requested):
Hi @Ragadeepika-Pucha and @stephjuneau , thank you for the notebook! A few comments:

- The NB needs a standard Authentication section (see Template).
- "This notebook is similar to the DESI EDR tutorial notebook, but it has been modified such that installing the DESI software (desispec) is not necessary to run this notebook." --> "[...] is not necessary to run it."
- "As a result, it can be run using a Python 3 kernel with only the Data Lab and SPARCL dependencies." --> "As a result, it can be run using the a Python 3 kernel with only the Data Lab and SPARCL dependencies (which are installed at Data Lab)."
- It's a bit hair splitting, but "DESI ... was built by the DOE" is not quire right, no? "Funded my the DOE" maybe? "Constructed under DOE contract"?
- Just a suggestion, but it might be good to mention both the English and the traditional names, e.g. "Kitt Peak / Ioligam, Arizona", or "Kitt Peak (Ioligam), Arizona"
- "~14,000 $deg^2$" --> "~14,000 $\rm deg^2$" (add \rm after the first $ sign and before 'deg')
- In the Imports cell: "## Data Lab related modules" --> "# Data Lab related modules". Or, probably you prefer: "# import some helpful python packages" --> "## import some helpful python packages"
- "## Load SPARCLClient" --> "## Instantiate SPARCLClient"
- "We focus on the zpix table in this notebook, which is the main redshift catalog from the DESI survey. It contains the redshift information from the coadded-spectra, based on the healpix of the objects. 2. target: This table contains targeting information of the different sources." --> The "2." before 'target' table seems out of place?
- "We look at the available columns in this tables below." --> "this table". Also, can skip the "below".
- "This can be cmx, special, sv1, sv2, or sv3." --> Decide if these will be always in italics or never. Currently there is a mix.
- "FIBER ASSIGNMENT PROGRAM. This is the planned observing conditions for the target." --> "These are the planned ..." ?
- "The different spectra observations of individual targets are coadded within each survey and program." --> Can you skip "observations" ?
- In section "Accessing DESI Targets" I feel that the first query cells needs a sentence of introduction. There are the comments, but below the query, and only as comments. Why not have that info in a nice Markdown cell?
- In the entire comment block, use small caps for all column names. Currently there is a mix of cases.
- "zcat_primary -- Whether or not a given coadded spectra is the primary spectra." --> "zcat_primary -- Whether or not a given coadded spectrum is the primary spectrum."
- "## desi_target -- encodeds main survey's desi targeting information ..." --> "encodes" and "DESI"
- In the first query, I understand why you do `CAST(zp.zcat_primary as int)` (because otherwise the values are strings 't' and 'f', and not usable for bit logic. But the CAST returns int64, i.e. 8 bytes, just to hold vales True or False (1 bit!).
This workaround doesn't work with Astropy's Table (or I don't know how; a simple `df['zcat_primary'].astype(bool)` doesn't work). But with `fmt='pandas'` in the `qc.query()` call, you could do:
```
query = """
SELECT zp.targetid, zp.zcat_primary
...

zpix = qc.query(sql = query, fmt = 'pandas')
zpix['zcat_primary'].astype(bool)
```
This saves a lot of memory.

- Remove everywhere the `%%time` cell directives. We have a built in timer now in JupyterLab.
- "Myers et al. 2023 describes the target selection" --> Technically, "describe".
- "there are also secondary fiber targets, that are based on different secondary programs in DESI" --> no comma
- "The information about different target bits are available here." --> "is available"
- "In the source of the markdown cell that contains the section "Selecting Sources based on Targeting Information", replace ` </br>` with an extra newline. (no need to involve HTML here.)

This also happens in other markdown cells.

- This block really wants a function:
```
## Manually check for the bit corresponding to the target classes
is_bgs = (sv1_desi_tgt & (2**60) != 0)|(sv2_desi_tgt & (2**60) != 0)|(sv3_desi_tgt & (2**60) != 0)
is_lrg = (sv1_desi_tgt & (2**0) != 0)|(sv2_desi_tgt & (2**0) != 0)|(sv3_desi_tgt & (2**0) != 0)
is_elg = (sv1_desi_tgt & (2**1) != 0)|(sv2_desi_tgt & (2**1) != 0)|(sv3_desi_tgt & (2**1) != 0)
is_qso = (sv1_desi_tgt & (2**2) != 0)|(sv2_desi_tgt & (2**2) != 0)|(sv3_desi_tgt & (2**2) != 0)
is_mws = (sv1_desi_tgt & (2**61) != 0)|(sv2_desi_tgt & (2**61) != 0)|(sv3_desi_tgt & (2**61) != 0)
is_scnd = (sv1_desi_tgt & (2**62) != 0)|(sv2_desi_tgt & (2**62) != 0)|(sv3_desi_tgt & (2**62) != 0)
```
Replace with, e.g.:
```
def bitcheck(bit):
    val = 2**bit
    res = (sv1_desi_tgt & val != 0)|(sv2_desi_tgt & val != 0)|(sv3_desi_tgt & val != 0)
    return res

is_bgs = bitcheck(60)
is_lrg = bitcheck(0)
is_elg = bitcheck(1)
is_qso = bitcheck(2)
is_mws = bitcheck(61)
is_scnd = bitcheck(62)
```
(and you can re-use the function later, in the cell that starts with `## MWS bit - 61`)

- This block is also very repetitive code:
```
fig, axs = plt.subplots(4, 1, figsize = (9, 12))
bins = np.arange(0, 4, 0.2)

axs[0].hist(zpix_cat['z'][is_bgs], color = 'C0', bins = bins, label = f'BGS: {n_bgs} sources')
axs[0].legend(fontsize = 14)
axs[0].set_ylabel("N(z)")
axs[1].hist(zpix_cat['z'][is_lrg], color = 'C1', bins = bins, label = f'LRG: {n_lrg} sources')
axs[1].legend(fontsize = 14)
axs[1].set_ylabel("N(z)")
axs[2].hist(zpix_cat['z'][is_elg], color = 'C2', bins = bins, label = f'ELG: {n_elg} sources')
axs[2].legend(fontsize = 14)
axs[2].set_ylabel("N(z)")
axs[3].hist(zpix_cat['z'][is_qso], color = 'C3', bins = bins, label = f'QSO: {n_qso} sources')
axs[3].legend(fontsize = 14)
axs[3].set_ylabel("N(z)")
axs[3].set_xlabel("Redshift")
```
Why not introduce a tiny function that does the per-panel work? E.g.:
```
fig, axs = plt.subplots(4, 1, figsize = (9, 12))
bins = np.arange(0, 4, 0.2)

def plotaxis(axid,mask,color,cat):
    axs[axid].hist(zpix_cat['z'][mask], color = color, bins = bins, label = f'{cat}: {n_bgs} sources')
    axs[axid].legend(fontsize = 14)
    axs[axid].set_ylabel("N(z)")
    
plotaxis(0,is_bgs,'C0','BGS')
plotaxis(1,is_lrg,'C1','LRG')
plotaxis(2,is_elg,'C2','ELG')
plotaxis(3,is_qso,'C3','QSO')
axs[3].set_xlabel("Redshift")
```
- "The SV*_DESI_TARGET column consists of ..." --> These are multiple columns, right? Then  "columns consist of"
- `jj = (zpix_cat['zcat_nspec'] > 3)&(zpix_cat['spectype'] == 'STAR')` --> Have spaces around the ampersand.
- Isn't random ;-)
```
## Randomly select an object

ii = 50
targetid = int(tsel['targetid'].data[ii])
```
- `targetid = int(tsel['targetid'].data[ii])   ## SPARCL accepts only python integers` --> I don't understand the inline comment.
- "# Plotting this spectra" --> "this spectrum" or "the spectra"
- "# Over-plotting smoothed spectra in black" --> "smooth spectrum" (singular)
- I think that spectrum plot should say whether the spectra are in rest frame or observed frame.
- "Let us now plot all the available spectra available for this object." --> Double "available".
-

Ragadeepika-Pucha:
@stephjuneau @rnikutta @jacquesalice @adambolton 

@rnikutta - Thank you for the comments. I have implemented most of them - there are a few things that I did not change, like the query comments and the plotting, to be consistent with the other notebook. 

- I did not change the query to "pandas" as well, because that requires changes in the next few cells as well. Cell number 12 was giving wrong result with pandas dataframe.
- Thank you for the suggestion about the bitcheck function - I added another parameter to work with the next steps as well.
- Regarding the randomness of selection of spectra -- the plotting functions are written such that if you modify the ii value, it will still work. I added that as a comment in the cell.
- targetid = int(tsel['targetid'].data[ii])   ## SPARCL accepts only python integers --> When the "int()" is removed and the targetid is given as the data from the table, SPARCL does not work. That is the reason for the comment.

Thank you!

jacquesalice (approved):

jacquesalice:
Thanks @Ragadeepika-Pucha ! 
@rnikutta : can you update your review to Approved so we can merge this pull request? It's currently blocked

rnikutta:
Great, many thanks @Ragadeepika-Pucha ! Just two small things:
* When one isn't logged in (like I just wasn't), you have to use the Auth cell. But for that to work, authClient must be imported. Could you please change this one line in the Imports cell to be:
```
from dl import queryClient as qc, authClient as ac
```
* You did not take my suggestion / code to use a small plot function, instead of repeating much code. Any reason?

rnikutta:
Hi @Ragadeepika-Pucha , as we discussed yesterday with @stephjuneau , I've added the authClient import myself, and will merge your PR in a second. Thank you for the NB!
In a future PR, we can modify the plotting code to define a small function (as suggested in the conversation above), but then make that change in both the DESI and Py3 kernel versions of the NB. It's definitely not urgent.

rnikutta (approved):

Ragadeepika-Pucha:
@rnikutta Thank you for merging the PR! I will add the plotting function to both the notebooks as soon as I can.

PR #191, opened 2023-08-15
Adaption of SPARCL+prospect notebook

jacquesalice:
This notebook is an adaption of the one created by Ben [here](https://github.com/desihub/prospect/blob/main/doc/nb/Prospect_spectrum_service.ipynb). There are no significant differences other than style and general layout.

The .ipynb file shows the prospect plot output only for DESI and not SDSS or BOSS due to file size restrictions. I can also clear the output from the DESI plot cell if need be.

NOTE that this notebook uses the DESI 23.1 kernel.

Lmk if I should make any changes! @weaverba137 @rnikutta


weaverba137 (commented):
Overall this looks good. However, there are a few things I'm concerned about.

1. This could be just how GitHub renders notebooks, but I can't see the sections on SDSS and (e)BOSS, or anything below where the DESI visualization is initiated.  Are those sections still there?  If so, it might be better to save this notebook without the outputs.
2. Please add a link to the original notebook in the prospect repository, maybe in the Disclaimer & attribution section.

jacquesalice:
@weaverba137 oh that's weird. Ok I just cleared the output for cells that show the plots and the SDSS and (e)BOSS sections are visible now. Also added the link to the original notebook.

weaverba137:
Thank you. I also note, for the record, that when the prospect/bokeh outputs are included, the notebook is about 13 MB.  Without the outputs, it is about 125 KB, so the outputs make the notebook 100x larger.

weaverba137 (approved):

PR #203, opened 2023-09-14
Fix a previous conflic around the __version__ string

rnikutta:


rnikutta:
Merged myself due to minimal nature of the PR.

PR #213, opened 2024-01-24
Fix fully-masked SDSS spectrum

weaverba137:
This PR adds a test for fully-masked spectra, which cause problems for current versions of Prospect. The test is added to DESI and BOSS spectra as well, for completeness.

This notebook also demonstrates how to hide only a particular class of warning that we don't care about.

Please test, then we'll update the HTML file, and add the cell outputs back into the notebook.

### DIFF
--- a/04_HowTos/SPARCL/Plot_SPARCL_Spectra_with_Prospect.ipynb
+++ b/04_HowTos/SPARCL/Plot_SPARCL_Spectra_with_Prospect.ipynb
@@ -2,20 +2,12 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 1,
-   "metadata": {
-    "execution": {
-     "iopub.execute_input": "2023-08-16T17:27:39.943059Z",
-     "iopub.status.busy": "2023-08-16T17:27:39.942752Z",
-     "iopub.status.idle": "2023-08-16T17:27:39.948412Z",
-     "shell.execute_reply": "2023-08-16T17:27:39.947306Z",
-     "shell.execute_reply.started": "2023-08-16T17:27:39.943021Z"
-    }
-   },
+   "execution_count": null,
+   "metadata": {},
    "outputs": [],
    "source": [
     "__author__ = 'Eric Armengaud, Benjamin Weaver <benjamin.weaver@noirlab.edu>, Alice Jacques <alice.jacques@noirlab.edu>'\n",
-    "__version__ = '20230816' # yyyymmdd\n",
+    "__version__ = '20240124' # yyyymmdd\n",
     "__datasets__ = ['sdss_dr16', 'boss_dr16', 'desi_edr']  \n",
     "__keywords__ = ['sparcl', 'spectroscopy', 'sdss spectra', 'desi spectra', 'tutorial', 'prospect', 'specutils']"
    ]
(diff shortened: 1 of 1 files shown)


stephjuneau (changes requested):
Hi @weaverba137 , thanks for modifying this notebook! I have a few comments below:

- While I see the authors include Eric, yourself and Alice in the top cell, I suggest also adding all 3 names under the title to make it more obvious.
- The html version will need to be made again after the NB is finalized
- I like how you check if any of the spectra are fully masked for all data sets. In the case where one spectrum is found to be masked, I suggest to save this information as a flag or boolean expression to apply to the Spectrum1D object later. For example I tried modifiying the cell like so: 
```
# initialize array with False and check on each spectrum below
fully_masked_sdss = np.array(np.zeros(len(sdss_spectra.records)), dtype='bool')

for k in range(len(sdss_spectra.records)):
    assert np.isfinite(sdss_spectra.records[k].flux).all()
    assert np.isfinite(sdss_spectra.records[k].ivar).all()
    if (sdss_spectra.records[k].mask > 0).all():
        print(f"WARNING: Spectrum record {k:d} is fully masked!")
        fully_masked_sdss[k] = True
```
And then later on I tested replacing the range of indices defined manually as `[1:,:]` by the expression `[~fully_masked_sdss,:]`, so the Spectrum1D cell would become: 
```
sdss_prospect = Spectrum1D(flux=flux[~fully_masked_sdss, :] * u.Unit('1e-17 erg / (Angstrom cm2 s)'),
                           spectral_axis=sdss_spectra.records[0].wavelength * u.Unit('Angstrom'),
                           uncertainty=InverseVariance(uncertainty[~fully_masked_sdss, :]),
                           mask=mask[~fully_masked_sdss, :] != 0,
                           meta=meta)
```

And similarly under the Model section:
```
sdss_model = (sdss_spectra.records[0].wavelength, model_flux[~fully_masked_sdss, :])
```
What do you think about implementing something similar as it would become more robust to the code being used (or copied) with other queries that a user might modify or define?

Happy to take a look at the next version. It's a great notebook to have!

weaverba137:
@stephjuneau, I've implemented your suggestions, although the HTML version remains to be generated.  I've applied the fully-masked check to all of DESI, SDSS, BOSS. Please test.

stephjuneau (commented):
Hi @weaverba137 , thanks for the revised notebook! I have a few comments:

1) I see that two notations are used for defining the subset of spectra/models to display. They seem equivalent as both "work" but we could consider homogenizing for stylistic consistency. For example: `[~desi_fully_masked, :]` versus `[~desi_fully_masked]` both seem to work in the same way when applied to, e.g., `flux` (unless this is only true when selecting everything; I haven't experimented extensively)

2) The Prospect window shows "other model" in dashed line for DESI (perhaps expected?) but also for SDSS (definitely not expected) as shown in the screen shot below. I am not sure where this comes from. The drop-down menu for the other model between best-fit, 2nd best-fit, etc. behaves partly consistently with expectations as in the std star, galaxy, etc. are fine and the best fit is the regular best-fit but it's not clear why there is an alternative model in dashed line originally.

3) Just noticed this but it may have been there before and be a Prospect issue but navigating to the Next/Previous spectrum with the arrows doesn't always update the slider and the number over the slider self-consistently. This could be in a separate investigation/PR as it doesn't impact being able to visualize the spectra but it could be confusing.

image

weaverba137:
> I see that two notations are used for defining the subset of spectra/models to display. They seem equivalent as both "work" but we could consider homogenizing for stylistic consistency. For example: [~desi_fully_masked, :] versus [~desi_fully_masked] both seem to work in the same way when applied to, e.g., flux (unless this is only true when selecting everything; I haven't experimented extensively)

Those objects have different dimensions, so the different notations are actually completely necessary.

> The Prospect window shows "other model" in dashed line for DESI (perhaps expected?) but also for SDSS

I will take a look at that.

> Just noticed this but it may have been there before and be a Prospect issue but navigating to the ...

I believe that is a Prospect issue outside the scope of this notebook.

stephjuneau:
> > I see that two notations are used for defining the subset of spectra/models to display. They seem equivalent as both "work" but we could consider homogenizing for stylistic consistency. For example: [~desi_fully_masked, :] versus [~desi_fully_masked] both seem to work in the same way when applied to, e.g., flux (unless this is only true when selecting everything; I haven't experimented extensively)
> 
> Those objects have different dimensions, so the different notations are actually completely necessary.

OK but the case that first caught my eye was:
`sdss_model = (sdss_spectra.records[0].wavelength, model_flux[~sdss_fully_masked, :])` having the same notation for boss but for desi:
`desi_model = (desi_spectra.records[0].wavelength, model_flux[~desi_fully_masked])` without the `,:`.
Is this expected?

> > The Prospect window shows "other model" in dashed line for DESI (perhaps expected?) but also for SDSS
> 
> I will take a look at that.

OK thanks!

> 
> > Just noticed this but it may have been there before and be a Prospect issue but navigating to the ...
> 
> I believe that is a Prospect issue outside the scope of this notebook.

Sure, could you create an issue for tracking this separately somewhere?

weaverba137:
> The Prospect window shows "other model" in dashed line for DESI (perhaps expected?) but also for SDSS
> `desi_model = (desi_spectra.records[0].wavelength, model_flux[~desi_fully_masked])`

Thanx, that was a typo. I've fixed that one.

> The Prospect window shows "other model" in dashed line for DESI (perhaps expected?) but also for SDSS

This is actually happening for all three of DESI, SDSS, & BOSS. It is showing an "other model" for a *different spectrum*, and it is probably related to the arrow-slider issue (item 3) identified above.

The arrow-slider synchronization appears to be working for every *other* spectrum. In between we get a "ghost" model for the last-displayed spectrum. If you look very carefully, and advance through the spectra, you should only see the "ghost" model on every other spectrum.

Before we submit a ticket, let's make sure gp12 has the latest version of Prospect. Currently it has 1.2.5, but 1.3.1 is the latest.  Since gp12 is due for imminent rebuilding, this could take a while. In the meantime, perhaps you can test on a system that has Prospect 1.3.1?

weaverba137:
I've created desihub/prospect#98.

stephjuneau:
OK, so I think that as far as this notebook is concerned, we could merge this version and separately follow the Prospect issue(s) such as in the new ticket you created. So I'll go ahead and approve this version now that you fixed the typo!

stephjuneau (approved):
Looks good in terms of the SDSS, BOSS and DESI coding. Some Prospect issues will be resolved elsewhere.

weaverba137:
I confirmed that the issue is still present in Prospect/1.3.1, so I don't think anyone else needs to try to re-confirm that.

rnikutta:
Thank you for the fix @weaverba137 and for the review @stephjuneau ! Merged.

PR #218, opened 2024-04-28
Update NB to run on gp13, where astropy.stats.LombScargle() is now as…

rnikutta:
…tropy.timeseries.LombScargle()


rnikutta:
@jacquesalice can you please review this NB PR on repgp02? Only change is where Astropy now has the LombScargle() class.

PR #219, opened 2024-05-01
Update EPO NB Breaking_the_Solar_System with new tqdm naming.

rnikutta:


rnikutta:
@mjfitzpatrick please review. Changed in `Files/tools.py` how the tqdm progress bar is imported. This gets rid of the DeprecationWarning. Note that on first use the progress bar does not appear to be working, but it works on second use. I suspect this is because in the first case, the steps are spread over multiple cells, while in the 2nd case they are all in one cell.

jacquesalice (approved):

PR #225, opened 2024-05-29
Update NB template: disclaimer and acks, available datasets section, …

rnikutta:
…commented out login cell.

### DIFF
--- a/DataLabNotebookTemplate.ipynb
+++ b/DataLabNotebookTemplate.ipynb
@@ -3,11 +3,20 @@
   {
    "cell_type": "code",
    "execution_count": 1,
-   "metadata": {},
+   "metadata": {
+    "execution": {
+     "iopub.execute_input": "2024-05-29T15:02:30.914526Z",
+     "iopub.status.busy": "2024-05-29T15:02:30.913739Z",
+     "iopub.status.idle": "2024-05-29T15:02:30.922053Z",
+     "shell.execute_reply": "2024-05-29T15:02:30.920340Z",
+     "shell.execute_reply.started": "2024-05-29T15:02:30.914467Z"
+    },
+    "tags": []
+   },
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


rnikutta:
@stephjuneau @jacquesalice @galaxyumi I've updated the DL NB template:
- Linked to our new Disclaimers page
- Added verbatim our detailed Acks requirements

I also edited the cell that lists all available datasets, and commented out the Login cell.

Please kindly review the template NB. The Disclaimer and Acks cell will be used to update all other NBs in the next step.

jacquesalice (approved):
Looks good to me, thanks Robert!

stephjuneau (commented):
Hi @rnikutta , the new acknowledgements all look good!

I know that not the main point of this PR but I was just curious about why commenting our the Authentication part? Shouldn't the line `ac.whoAmI()` just result in "anonymous" when the user is not authenticated, which would be useful to know for the user in any scenario?

rnikutta:
It's a good point.. The reason I commented it out is because in most cases (I think) a user running a NB will be logged in, including when we test the NBs ourselves (which then always evaluates to my username...). To avoid having to explicitly use ac.logout(), then remove that code again, and re-run the NB, I thought it was easier to comment out the cell. The text just before the auth cell actually says explicitly to un-comment the cell, should a login be needed.

What do you think?

jacquesalice:
> It's a good point.. The reason I commented it out is because in most cases (I think) a user running a NB will be logged in, including when we test the NBs ourselves (which then always evaluates to my username...). To avoid having to explicitly use ac.logout(), then remove that code again, and re-run the NB, I thought it was easier to comment out the cell. The text just before the auth cell actually says explicitly to un-comment the cell, should a login be needed.
> 
> What do you think?

If we comment it out in this notebook, should we comment it out in all notebooks?

rnikutta:
I've been suggesting as much in most of the NBs that I've been reviewing over the past months..

stephjuneau (approved):

rnikutta:
Thank you, will merge!

PR #244, opened 2024-08-01
Work-around for pyraf import setting CWD to a user's home dir.

rnikutta:
@bmerino95 please review and approve this PR.


bmerino95 (approved):
I just tried this and verified that the updated notebook downloads the files to my working directory instead of /home.

PR #250, opened 2024-09-20
Update File Services NB: SDSS DR14->DR16, add TOC, streamline and mod…

rnikutta:
…ernize a bit.


rnikutta:
@weaverba137 can you please kindly review the small update to this NB? Thank you.

weaverba137:
Thank you, I will take a look when I get a chance.

weaverba137 (approved):
I ran the notebook on `gp13` with no errors, so that's good.

I noticed a couple of style quirks where there is a space between a print statement and the function parentheses: `print (`.  That's not technically incorrect, but a Python style analyzer would probably complain about that.  It also makes the code look more like Python 2 than Python 3. There are also 1-2 places with `sc.ls (`.

So just a minor style quibble!

rnikutta:
@weaverba137 I did the stylistic fixes, if you want to take another quick look? Thank you.

weaverba137:
Looks good, go ahead and merge.

PR #259, opened 2025-01-29
Updated table of contents and markdown cells for GHOST notebook.

bmerino95:
I have shortened the table of contents for the GHOST notebook and updated some markdown cells.


jacquesalice (changes requested):
Hi @bmerino95 . The notebook looks good! Just a few things:
1. I was wondering if there was a reason you link to v3.2.1 of the DRAGONS readthedocs page instead of the stable version in your Summary section?
2. In the Summary section, change "...when a DRAGON command..." --> "when a DRAGONS command..."
3. Also in the Summary section, change "In this notebook, we present an example of a DRAGONS Jupyter notebook..." --> "We present an example of a DRAGONS Jupyter notebook..." or something like that
4. Somewhere in the Summary section I think you should mention the maximum amount of storage space the downloaded and altered files will [temporarily] take up. In case someone would be worried about that. From what I see, it's no more than ~3 GB at any point.
5. In the bias reduction section, the filenames you say will be created are different from what is actually created:
S20230417S0040_slit_bias.fits .. instead S20230417S0011_slit_bias.fits
S20230417S0012_red001_bias.fits .. instead S20230417S0011_red001_bias.fits
S20230417S0038_red001_bias.fits .. instead S20230417S0036_red001_bias.fits
S20230417S0039_blue001_bias.fits .. instead S20230417S0036_blue001_bias.fits
6. And in the flats section:
S20230416S0047_red002_flat.fits .. instead S20230416S0047_red001_flat.fits

bmerino95:
Hi @jacquesalice. Thank you for the feedback! I don't know how I missed so many file names being mislabeled. I just went through and made all the changes you suggested.

jacquesalice (approved):
Thanks @bmerino95 ! Approving

PR #267, opened 2025-03-19
updated Py3 notebook and small fixes to markdown in dr1

stephjuneau:
- Modified the Py3 version to DR1 and to be nearly identical to the `01_Intro_to_DESI_DR1.ipynb` notebook

- Made small fixes to some markdown comments in the 01_Intro_to_DESI_DR1 notebook as I was going through it.

I think this concludes revising all DESI notebooks (except for the small renaming conflict for 02_ notebook that we can do after merging to main).


weaverba137 (approved):
Looks good to me, merge when ready.

PR #269, opened 2025-03-19
Sj fix desi nbs

stephjuneau:
- Renamed the 02_DESI_SDSS_Comparison nb
- Applied the same small fixes to the query on nspec>5 to remove spectype and add ZWARN label on plots with an explainer below (now they're both consistent with the DESI team version)


stephjuneau:
@jacquesalice -- this is ready for review when you have a moment!

PR #274, opened 2025-05-19
ipyaladin example notebooks

bmerino95:
I have created two example notebooks that are ready for testing. The first one, ipyaladin_moc.ipynb gathers data from three surveys, generates MOCs for them, and displays the MOCs in the Aladin Lite viewer. The second one, ipyaladin_globular_cluster.ipynb, uses the SIA tool to download images of the globular cluster NGC 1851 and overlays them onto Aladin. After that, the notebook pulls some survey data to plot individual data points onto the viewer. 

Both notebooks use a Jupyter Notebook module called Sidecar that displays the Aladin Viewer on the right-hand side. I have not experienced any issues with Sidecar, but please let me know if you encounter anything strange or if Aladin isn't appearing. 

If you have the time, can you test these notebooks out?


rnikutta:
@bmerino95 seems like you accidentally added an entire Python environment with all its files to this PR. Closing. Please make a new, clean one.

PR #275, opened 2025-05-20
ipyaladin example notebooks v2

bmerino95:
Hi @rnikutta! Sorry about that. My last PR attempt contained a folder and a Python environment where I tested some things from the ADASS talk. I have removed them and a different test_space directory, so this PR should be ready for review. From my side, things look better, but please let me know if you see anything besides the two notebooks, a readme, and a PNG file. 

Thank you!


rnikutta:
@bmerino95 here my comments for the other NB (globular cluster)

- Change the date format of the `__version__` variable to (as indicated) yyymmdd.

- Change 'Aladin' keyword to 'AladinLite'; maybe add 'ipyaladin'.

- In Goals and/or Summary sections, change the focus. Instead of

"Demonstrate the process of querying the Data Lab utilizing the query client service and the Simple Image Access (SIA) protocol to retrieve image cutouts. Show how to display the catalog and image data using the Aladin Lite viewer."

write for instance

"Demonstrate how to display image and catalog data with the AladinLite Viewer in a notebook (using ipyaladin). The image and catalog data will be obtained through Data Lab's data services."

- In Goals and/or Summary sections, add links to the AladinLite viewer website. Also to ipyaladin (if there is such a site).

- The Summary section speaks of MOCs, but those were featured in the other NB (ipyaladin_MOCs.ipynb). Update please.

- Update the "Disclaimer & Attribution" section to the newer long-form version (see DataLabNotebookTemplate.ipynb; or any other existing NB).

- Reorder the imports: std lib first, 3rd party next, DL last.

- Both the SideCar and AladinLite instance accept a few arguments that you could use to open the viewer larger and in "full screen" (meaning "full panel") size:

  ```
  aladin = Aladin(full_screen=True)
  with Sidecar(title="aladin_output",anchor='split-right'):
      display(aladin)
  ```

- Add this line after `aladin.target = target_coord`:

  ```aladin.fov = 0.5  # set field of view to 0.5 degrees```

- Immediately after after this, write a few words how the user can pan/zoom/change layers, etc., using the mouse on the right hand side.

- In the SIA section, the number of rows in the image table pre and post applying the prodtype and proctype filters is the same (33). Not a good example maybe? (can't leave it as-is, since it appears a pointless exercise).

- From the final image table, you select "first, fourth, and seventh" images. Why? This must be motivated.

- The code for selecting these images, downloading them, creating the WCS objects, etc., is almost identical. Write a short function that takes the image table and a number as arguments, and returns whatever you need for plotting.

- Similarly, for the 3-panel plot, label the panels with something more meaningful than "Image 1" (e.g., the filter band?). Spend a sentence or two explaining what we are seeing.

- You are extracting the WCS of the images, but you aren't making use of them in the plots. If you use the correct projection (passing the WCS), you get plot axes nicely labeled with RA and DEC, rather then the current "pixel number".

- When I'm trying to overlay the images in Aladin (on gp13 NB server), I get this error:

```
aladin.add_fits(url_1,name='image_1',opacity=1.0,colormap='red')

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[64], line 2
      1 #Add Image_1 to Aladin
----> 2 aladin.add_fits(url_1,name='image_1',opacity=1.0,colormap='red')
      4 #Add image_2 to Aladin
      5 aladin.add_fits(url_2,name='image_2',opacity=0.5,colormap='green')

AttributeError: 'Aladin' object has no attribute 'add_fits'
```

Is this supposed to work with the AladinLite version that's installed on gp13?

- "let's query the data lab ..." --> "Data Lab"

- It's a very cool and common usecase to overlay source positions from a catalog over an image! But, doing this in your example, I see two issues currently:

  1) There are still too many sources --> visual blob. Can you maybe limit the DB query to for instance only stars brighter than a certain magnitude? (also, I see nothing in your query that would limit the query to "likely stars"! You want to exclude any potential extragalactic background sources). 
  2) I don't see many circles actually overlapping with the sources in the image. This needs investigation. Could be the astrometry of the image. Or of the catalog. Could be issues with reference coordinate frames? etc.

- Once the catalog sources are overplotted, you need some text to explain stuff. What do we see? Maybe you want to focus on bright stars (as I suggested above), so point them out (once you get the misalignment sorted out).

rnikutta:
@bmerino95 here my additional comments regarding the catalog query and overlay section:

- "let's query the data lab to get data for individual stars in and around our target" --> "let's query Data Lab catalogs to get photometry for individual stars in and around our target"

- "We will use the SkyMapper data release 4 table for this notebook because our target falls within the survey's footprint." --> "We will use SkyMapper data release 4."

- "Since we already know the coordinates of our target, we can perform a cone search centered around the globular cluster. Note that we set the FOV to be 0.2 degrees when working with images because we wanted to find any images around our target, so we cast a wide net. For this query, we will need to narrow our FOV because SkyMapper DR4 contains data for many sources near our target. Using the same FOV as the image search would return 44,470 sources! Many of which have nothing to do with our target. If we narrow our FOV by just a factor of 3, the radius of the cone search becomes about twice as large as the diameter of the globular cluster. This query will still return a lot of sources, but this time, most of them are closer to our target."

--> This is too long and mostly unnecessary. Just quote the angular diameter of the GC from literature and say you going to do a cone search of that diameter. Also "centered on the globular cluster", not "around the globular cluster".

- You are querying the skymapper_dr4.photometry table, which is the measurements table. Query the master table (which is the object table) instead. Decide on which filter bandpass X you want to use (it should match the image bandpass that you are displaying), and then query for X_psf. Also use raj2000 and dej2000 (the mean ra and dec of an object).

- If you want to apply a magnitude cut, I propose you include the magnitude cut in the SQL query already. No point in fetching 6k sources just to plot a few hundred. Explain before the query why you are using a magnitude cut.

- "The SkyMapper survey contains tens of thousands of data points." --> add "within our search cone."

- "the image would be obscured by a blanket of circular regions" --> "the image would be obscured by a blanket of data markers"

- "only sources with magnitudes less than 16 are displayed" --> "only sources with magnitudes brighter than 16 are displayed"

- No need to show the overlaying of catalog object both with matplotlib and with ipyaladin. Only show with ipyaladin.


- I have an idea for an additional plot, which would make the notebook a lot more science-y: query for stars in the GC (as many as you can reliably get), and then plot a color-magnitude diagram of their photometry. It's a famous exercise to determine the age of a globular cluster from the color-magnitude diagram of its stars, by identifying the main-sequence turn-off point.

rnikutta:
Hi @bmerino95 , reviewing your latest updates:

- We usually write the author(s) of a NB also under the the title (in italics). Compare with existing NBs.
- Nice overlay of circles. Now most have a nice dot in them :)
- Do we understand why despite your small adjustment of he GC's coordinates, the three images in the matplotlib figure are still offset from the center?
- On the CMD:
  - I don't understand this: "instead of using a cone search with a cone radius equal to the angular radius reported in literature, we will divide the FOV by 2 instead. Widening the cone search will increase the number of data points for the CMD, allowing us to identify key characteristics of the diagram." --> If you are dividing the FOV by 2, your search angle shrinks. But you said you want more data points, not fewer()?)
  - Embrace vectorized operations and that you can just operate on entire Pandas columns :) Remove the entire cell that computes the g-i color, and just plot the CMD directly like this:
  ```
  plt.scatter(df2['g_psf']-df2['i_psf'], df2['i_psf'], s=.5)
  ```

rnikutta:
Hi @bmerino95 I finally checked your latest changes, thank you! Just three small requests, and then we can merge:
- When you display the z band image, while it aligns correctly with the previous view, both are de-centered. Can you add a line of code to re-center the view, such that the center of the GC aligns with the haircross?
- It's u,g,r,i,z,Y band, i.e., small caps except for Y-band. Check Skymapper and DES websites. Fix everywhere (including in figure axes labels).
- At the very end, the text under the CMD figure schematically explains what isochrone matching can do, but stops there. We discussed that we don't have to do the isochrome matching in this NB. But then at least add a sentence "We won't perform isochrone matching in this notebook." (or something similar).

bmerino95:
Thank you for the feedback @rnikutta! I removed the empty cell at the end of the notebook and kept the outputs this time.

rnikutta (approved):
Thank you @bmerino95 looks great! Merging.

kareninysimba:
Hi @bmerino95 ,
When running the "Imports" cell of the `ipyaladin_MOC.ipynb` notebook, in gp13, I'm getting 

> ModuleNotFoundError                       Traceback (most recent call last)
> Cell In[3], line 14
>      11 from astropy.table import QTable
>      12 from astropy.coordinates import SkyCoord
> ---> 14 from mocpy import MOC
>      15 from ipyaladin import Aladin
>      16 from sidecar import Sidecar
> 
> ModuleNotFoundError: No module named 'mocpy'
> 

Is there a particular kernel I have to choose different from Python 3 (DL,Py3.10.13) to run it?

rnikutta:
@bmerino95 is it possible that you and I installed locally a version of mocpy? When I load the NB @kareninysimba mentions on gp13 and do

```
import mocpy
mocpy.__file__
```

I get a local path:

```'/home/jail/dlusers/robertdemo/.local/lib/python3.10/site-packages/mocpy/__init__.py'```

If that's the case for all, then we need to ask @chaduhduh to install mocpy in the stack on gp13. (in addition to the needed jdaviz and Python updates...)

PR #278, opened 2025-07-10
GSAOI DRAGONS update

bmerino95:
The current GSAOI notebook fails to reduce the list of science images when using the new DRAGONS v4.0.0 kernel. The fix involved replacing the following line 

`reduce_target.uparms.append(('skyCorrect:offset_sky', False))`

with

`reduce_target.uparms['skyCorrect:offset_sky'] = False`

I have verified that the new line of code works on GP13 with the new kernel. I will note, however, that the fix does not work with the current DRAGONS kernel. Until the new kernel is published, this PR will remain as a draft.

### DIFF
--- a/04_HowTos/DataReduction/DRAGONS_reduction_examples/GSAOI_Imaging_EllipticalGalaxy/GSAOI_Imaging_EllipticalGalaxy.html
+++ b/04_HowTos/DataReduction/DRAGONS_reduction_examples/GSAOI_Imaging_EllipticalGalaxy/GSAOI_Imaging_EllipticalGalaxy.html
@@ -7312,6 +7312,7 @@ a.anchor-link {
                     processEnvironments: true
                 },
                 displayAlign: 'center',
+                messageStyle: 'none',
                 CommonHTML: {
                     linebreaks: {
                     automatic: true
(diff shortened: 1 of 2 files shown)


bmerino95:
Awesome! Thank you, @chaduhduh.

Hi @kareninysimba. Now that the new kernel is live, my updated GSAOI notebook is ready for review. I only changed one line, but could you run it to ensure it doesn't generate any errors?

kareninysimba:
Hi @bmerino95 , @chaduhduh 
When running the new version of the notebook for the 1st time, as I mentioned in our DataLab group meeting (7/21), the file downloading was too slow (around 1 file downloaded every 2-3 mins) and then (in 2 occasions) the download stopped with a message I've never seen (I copied it but I lost it when closing the browser accidentally). The result was that some files were not downloaded or had 0 bytes. This may have been a temporary glitch related to the archive more than the notebook or gp13.
It didn't happen again, but wanted to warn it here.

Notebook ran uneventfully today and as fast as it should in my latest run, using the (default) DRAGONS 4.0 kernel.
Zero issues.

kareninysimba (approved):
Temporary glitch during file download. Most likely unrelated to the notebook or the new kernel.
It ran uneventfully after that. Seems good.

PR #279, opened 2025-07-14
Replace GOGREEN DR1 with DR2

jacquesalice:
This PR replaces the use of GOGREEN DR1 with GOGREEN DR2 in 5 notebooks. This is at the request of Michael Balogh (P.I. of GOGREEN/GCLASS):

> DR2 completely supersedes DR1 and there is no reason for people to continue to use DR1.  So while the DR1 dataset should remain available "just in case", it would be best if all tools referenced only DR2, and it should be difficult for a user to "accidentally" access DR1. Therefore I think it's best to remove the old notebooks and just keep the DR2 versions.

@rnikutta can you please review?


rnikutta (changes requested):
Thanks @jacquesalice . I ran all the new NBs, confirm that all run fine and completely. Didn't find any missed references to DR1. Some other comments (especially the initial ones matter):

**GENERAL COMMENTS**

- To retire DR1 we need to announce it on the website, give a date (can be as short as a month?), then retire it.

- Together with retirement of DR1, the landing page needs to be updated.

- The dropping of DR1 does not need to mean that DR2 is not available (it already 
is!). But the DR1 NBs should not disappear until DR1 data are retired.

**NB 0**

The printouts with wrapper.fill() and without look like this. I.e., worse with the wrapper. What's the use of `wrapper.fill()`?

print_wrapper

**NB 1**

- Might remove the commented-out line `#MEF = fits.open(onedfile)` from function `recalcSNR()`.

- Having the '.' between schema and table names as part of "dataset" is so ugly ;-)

```
dataset = 'gogreen_dr2.' # <-- this dot!
tables = ['clusters', 'redshift']

for table in tables:
    globals()['%s_table'%table] = qc.query('select * from %s%s' % (dataset, table), fmt = 'pandas')
    print("%s table" %table)
    display_html(globals()[('%s_table'%table)].head(5))
```

Move it to the code in the loop and the SQL string instead.
(Same in other NBs, e.g. NB 4_*ipynb)

**NB 4:**

- The note at the end:

"Note:

If the [O II] equivalent width (ewoii) is 0 and results in a percent error of infinity, this means there is an error from the database (known issue for GOGREEN DR1, which has been fixed for GOGREEN DR2). Therefore, if using the DR1 version, use the ewoii_model to compare the estimated equivalent width to the number in the database.

..."

Do we need to keep it, if we are deprecating DR1? (maybe yes, since this is possibly the single place where the bug in DR1 is actually recorded?...)

jacquesalice:
@rnikutta thank you for reviewing! We plan to keep both GOGREEN DR1 and DR2 in our database, but only use DR2 in the notebooks and tell users to only use the DR2 version for their science. So I've added this statement to the beginning of each notebook and also to the README:
> **Note:** Astro Data Lab also hosts GOGREEN DR1. However, GOGREEN DR2 supersedes the DR1 version and should be used for any and all science cases.

I've implemented your other comments as well.

rnikutta (approved):
All good, thanks @jacquesalice ! Will merge in a sec.

Issue #194, opened 2023-08-24
Introduce a function for plotting z histograms in DESI EDR notebooks


stephjuneau:
Suggestion from @rnikutta  for a future revision of both DESI EDR tutorial notebooks. I'm marking it as en enhancement.

"This block is also very repetitive code:"

```
fig, axs = plt.subplots(4, 1, figsize = (9, 12))
bins = np.arange(0, 4, 0.2)

axs[0].hist(zpix_cat['z'][is_bgs], color = 'C0', bins = bins, label = f'BGS: {n_bgs} sources')
axs[0].legend(fontsize = 14)
axs[0].set_ylabel("N(z)")
axs[1].hist(zpix_cat['z'][is_lrg], color = 'C1', bins = bins, label = f'LRG: {n_lrg} sources')
axs[1].legend(fontsize = 14)
axs[1].set_ylabel("N(z)")
axs[2].hist(zpix_cat['z'][is_elg], color = 'C2', bins = bins, label = f'ELG: {n_elg} sources')
axs[2].legend(fontsize = 14)
axs[2].set_ylabel("N(z)")
axs[3].hist(zpix_cat['z'][is_qso], color = 'C3', bins = bins, label = f'QSO: {n_qso} sources')
axs[3].legend(fontsize = 14)
axs[3].set_ylabel("N(z)")
axs[3].set_xlabel("Redshift")
```

Why not introduce a tiny function that does the per-panel work? E.g.:

```
fig, axs = plt.subplots(4, 1, figsize = (9, 12))
bins = np.arange(0, 4, 0.2)

def plotaxis(axid,mask,color,cat):
    axs[axid].hist(zpix_cat['z'][mask], color = color, bins = bins, label = f'{cat}: {n_bgs} sources')
    axs[axid].legend(fontsize = 14)
    axs[axid].set_ylabel("N(z)")
    
plotaxis(0,is_bgs,'C0','BGS')
plotaxis(1,is_lrg,'C1','LRG')
plotaxis(2,is_elg,'C2','ELG')
plotaxis(3,is_qso,'C3','QSO')
axs[3].set_xlabel("Redshift")
```

PR #280, opened 2025-07-22
Updated DESI Notebooks 

Ragadeepika-Pucha:
This is linked to a closed [PR](https://github.com/astro-datalab/notebooks-latest/pull/207#issuecomment-3097871036).

Updated the notebooks based on the current versions. Main changes include:

- Updated my affiliation and contact information.
- Added the authentication cell to all the notebooks, to be consistent and to help with the queryClient.
- Added the reference links related to DESI DR1 paper.

With respect to the "enhancement" [issue](https://github.com/astro-datalab/notebooks-latest/issues/194) - I modified the cell to include a for-loop over the targeting classes instead of a plotting function. This is consistent across all the notebooks. Personally, I think that issue can be closed.

Let me know if anything else is needed/changed.
Thank you!


rnikutta (changes requested):
See my review comments in the post above.

Ragadeepika-Pucha:
Hi @rnikutta - I updated the author list and authentication in all notebooks. Thank you for the comments!

rnikutta:
Thank you @Ragadeepika-Pucha ! Just two minor things (sorry I didn't flag them yesterday):

- In NB "01 EDR", have all imports in a single cell (with comments as needed).
- In all NBs: in the Imports cells (but maybe also in other cells), sometimes a comment is introduced with a single '#' symbol, sometimes with a double '##'. For instance, in NB XX:
```
# DESI targeting masks - 
[...]

## DataLab related modules
```
Use a single '# ' (with white space right after it). Also, in this example above, remove the trailing stuff: `# DESI targeting masks - ` --> `# DESI targeting masks`

Thank you.

stephjuneau (approved):
Looks good, @Ragadeepika-Pucha !

@rnikutta , regarding your general comments:

- We decided to keep one EDR notebook until we fully deprecate/remove the desi_edr dataset (and updated most other notebooks with DR1).  I think we can start planning now whether (and when) we want to remove it and post a warning to users. We did add this note at the top of the EDR notebook: "Important: DESI DR1 is now available and supersedes DESI EDR; please refer to 01_Intro_to_DESI_DR1.ipynb" but we were leaving it available for people with work in progress using EDR or to replicate EDR results.

- Regarding the 01a version, it is intended for users who wish to run the notebook entirely locally without access to a DESI kernel and by request. So I'm leaning toward leaving it separately as it fulfills a specific use case and to avoid having to repeatedly include two scenarios in the notebook foe "DESI kernel" vs. "Py 3 kernel". Users are expected to use one or the other but would not need both. Happy to keep discussing these points though and I agree that they should not hold the PR from being merged.

Ragadeepika-Pucha:
Hi @rnikutta  - updated the notebooks with the suggested changes. Thanks!

rnikutta (approved):
Thank you @Ragadeepika-Pucha , looking good. Will merge in a second.

PR #281, opened 2025-07-22
GNIRS L-band longslit point source notebook

bmerino95:
With the release of DRAGONS v4.0.0, the DRAGONS team published four new tutorials for GNIRS longslit data. Since the L-band is the most requested band, in the past 5 years, I have created a Jupyter Notebook version of the tutorial. 

A highlight of the new tutorials is an interactive feature, but it does not work when using Jupyter Notebooks, so the Notebook I created only covers a general reduction using the default parameters. 

I have tested this notebook on GP13 (using the new kernel DRAGONS v4.0.0) and did not find any issues.


bmerino95:
Hi @kareninysimba! I have finished working on the new GNIRS L band longslit notebook. If you have time, could you try running it and let me know if you encounter any errors?

I have tested the notebook with the DRAGONS-4.0.0 (DL,Py3.12) kernel, and the only issue I have come across is the NFS file issue that seems to only occur intermittently.

PR #288, opened 2025-09-08
Major updates to HIPASS interactive NB

jacquesalice:
Firstly, this updated version of the HIPASS_interactive notebook fixes the error reported in [DLC-2116](https://noirlab.atlassian.net/browse/DLC-2116):
This (in a for loop) `I = [float(i) for i in Int]` fails at iteration 6/10 with "ValueError: could not convert string to float: ''"

Additionally, the following changes have been made to make the notebook better:
- Cleaner extraction of the data in the ASCII file
- Uses the SIA service to obtain cutout image preview URLs to use in the Bokeh plot (old version used SkyView from Astroquery to obtain cutout images, saved to VOSpace, then tried to use them in the Bokeh plot but they failed to show in the plot since Bokeh requires images used in the HoverTool to be URLs, not local files)
- Does not save anything to VOSpace

@rnikutta can you please review? The only part I'm not sure whether we should keep is the plot of all spectra in the "Extracting spectral information from the HIPASS database" section.


rnikutta (changes requested):
Thank you @jacquesalice , many nice improvements. Below I suggest some more:

- Fix TOC: headings and subheadings have inconsistent sizes (for their level)

- In the cell under "Plot the Sky coverage of the HIPASS survey" replace this

  ```
  ax.set_xlabel(r'$\mathrm{RA[\degree]}$')
  ax.xaxis.label.set_fontsize(22)
  ax.set_ylabel(r'$\mathrm{Dec[\degree]}$')
  ax.yaxis.label.set_fontsize(22)
  ```

  with this

  ```
  ax.set_xlabel(r'$\mathrm{RA[\degree]}$',fontsize=22)
  ax.set_ylabel(r'$\mathrm{Dec[\degree]}$',fontsize=22)
  ```

- The whole two-cell business "Choose dataset to visualise" is unnecessarily complicated. Just remove the entire cell starting with `# Report an error if 'selected' is wrong`.

- Under "Scraping url-s where the data of the HIPASS spectra is stored", replace this:

  `for galaxy in tqdm(range(df.index[0], df.index[0]+len(df))):`

  with the much simpler:

  `for galaxy in tqdm(range(len(df))):`

  And fix the writing in the heading to "Scraping URLs ...". Fix it also in the first comment line in the code cell.

- In the same cell, replace this logic:

  ```
  if len(str(df['cube'][galaxy]))==1:
      cube = ('00'+ str(df['cube'][galaxy]))
  elif len(str(df['cube'][galaxy]))==2:
      cube = ('0'+ str(df['cube'][galaxy]))
  else:
      cube = (str(df['cube'][galaxy]))
  ```

  with this:

  `cube = "%03d" % df['cube'][galaxy]`

- The business with creating the ascii1 and png1 URLs is a bit dirty. The clean way is to use urllib to do the url encoding. For instance like this:

 ```
 # just for one example
 from urllib.parse import urlencode # in the imports cell

 galaxy = 0
 cube = "%03d" % df['cube'][galaxy]
 ra = df['RAJ2000'][galaxy].decode().replace(' ',':')
 dec = df['DEJ2000'][galaxy].decode().replace(' ',':')
 xmin = 2000
 xmax = 12000

 root = 'https://www.parkes.atnf.csiro.au/cgi-bin/multi/release/download.cgi?'

 params = {"cubename":"/var/www/vhosts/www.parkes.atnf.csiro.au/htdocs/research/multibeam/release/MULTI_3_HIDE/PUBLIC/H%s_abcde_luther.FELO.imbin.vrd" % cube,
             "hann":3,
             "coord":"%s,%s" % (ra,dec),
             "xrange":"%s,%s" % (xmin,xmax),
             "xaxis":"optical",
             "datasource":"hipass",
             "type":"ascii"}

 ascii1 = root + urlencode(params,safe='/')
 params["type"] = "png"
 png1 = root + urlencode(params,safe='/')
 ```

- Under "Creating list of HIPASS sources" replace the entire code with just this:

 ```HIPASS_sources = ['HIPASS'+df['HIPASS'][idx].decode().strip() for idx in range(len(df))]```

- The cell that loads the spectra as ASCII tables from the CSIRO server doesn't catch errors, but 2/10 spectra throw errors. In addition, the CSIRO server returns status_code = 200 (meaning 'OK') even where there was an error. Finally, there were many inefficient and too verbose code snippets. Replace the entire cell that starts with

 ```
 cnt = 0
 fig, ax = plt.subplots(figsize=(8,6))
 ```

 with this instead:

 ```
 fig, ax = plt.subplots(figsize=(8,6))

 for idx,galaxy in tqdm(enumerate(all_ascii)):
     res = requests.get(galaxy)

     # The CSIRO server unfortunately returns res.status_code = 200 (meaning 'OK') even when there was an error.
     # Instead, we will check whether a phrase about an error is return in res.text
     if res.status_code == 200 and 'following error has occured' not in res.text:
         soup = str(BeautifulSoup(res.text,'html.parser'))

         # The spectrum is returned an ASCII table, as part of the server response.
         # `find` gets the table from the response string, no matter where it starts.
         # `+3` gets rid of the leading comment char.
         # `replace` fixes the column name with whitespace
         datastr = soup[soup.find('#  Channel')+3:].replace('Spectral axis','Velocity')
         df_spec = pd.read_csv(StringIO(datastr),delim_whitespace=True)
    
         ax.plot(df_spec['Velocity'], df_spec['Intensity'], label=f'{HIPASS_sources[idx]}')

 ax.set_xlim(2000,12000)
 ax.set_xlabel('VOPT (km/s)')
 ax.set_ylabel('Flux density (Jy beam^-1)')
 ax.legend(bbox_to_anchor=(1.05, 1))
 plt.show()
 ```

 and add `from io import StringIO` to the Imports cell.

- I'm not sure what to make of the SIA/cutout section. In no case do we see a clear optical counterpart galaxy. I checked for a few HIPASS sources with the Legacy Viewer, also nothing (but I didn't check them all). Hoever, paper III in the HIPASS series shows clear optical counterpart galaxies: https://ui.adsabs.harvard.edu/abs/2005MNRAS.361...34D/abstract
(check their figures 4 and 5).

jacquesalice:
@rnikutta Thank you for the review! I've implemented all of your suggested changes. As for your last comment on not seeing a clear optical counterpart galaxy, it's definitely more obvious in some images than others. Below are side-by-side comparisons of images from our SIA service vs. images from astroquery's Sky View (which is what the notebook used originally). Just to prove that it doesn't seem to be the fault of our SIA service.

HIPASSJ0656-36
HIPASSJ0656-36 hipass_image_HIPASSJ0656-36

HIPASSJ0836-43
HIPASSJ0836-43 hipass_image_HIPASSJ0836-43

HIPASSJ1009-12
HIPASSJ1009-12 hipass_image_HIPASSJ1009-12

HIPASSJ1339-31
HIPASSJ1339-31 hipass_image_HIPASSJ1339-31

HIPASSJ1756-05
HIPASSJ1756-05 hipass_image_HIPASSJ1756-05

HIPASSJ1816-49
HIPASSJ1816-49 hipass_image_HIPASSJ1816-49

HIPASSJ1931-06
HIPASSJ1931-06 hipass_image_HIPASSJ1931-06

HIPASSJ1935+01
HIPASSJ1935+01 hipass_image_HIPASSJ1935+01

HIPASSJ2057-43
HIPASSJ2057-43 hipass_image_HIPASSJ2057-43

HIPASSJ2123-69
HIPASSJ2123-69 hipass_image_HIPASSJ2123-69

rnikutta:
Hi @jacquesalice thanks for implementing the comments. Looks and works great. For many (if not all) of the sources there is now a clear galaxy in the cutout! Please feel free to merge.

PR #289, opened 2025-09-12
DRAGONS notebook maintenance

bmerino95:
Updated the "adapted from" and DRAGONS homepage links, changed the DRAGONS kernel to the latest version DRAGONS4.0.0 (DL,Py3.12), and added bad pixel masks to the download data section. When the notebooks were initially created, most of them did not require a bad pixel mask because they were handled differently; however, now they are handled as calibrations and need to be added to caldb manually. 

GMOS_Imaging_Galaxy.ipynb no longer exists on the DRAGONS readthedocs page, so its link remains unchanged.

### DIFF
--- a/04_HowTos/DataReduction/DRAGONS_reduction_examples/GMOS_longslit_WhiteDwarf/GMOS_longslit_WhiteDwarf.html
+++ b/04_HowTos/DataReduction/DRAGONS_reduction_examples/GMOS_longslit_WhiteDwarf/GMOS_longslit_WhiteDwarf.html
@@ -7805,7 +7805,7 @@ a.anchor-link {
 <div class="cm-editor cm-s-jupyter">
 <div class="highlight hl-ipython3"><pre><span></span><span class="n">__nbid__</span> <span class="o">=</span> <span class="s1">'0041'</span>
 <span class="n">__author__</span> <span class="o">=</span> <span class="s1">'Brian Merino &lt;brian.merino@noirlab.edu&gt;, Vinicius Placco &lt;vinicius.placco@noirlab.edu&gt;'</span>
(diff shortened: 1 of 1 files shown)


rnikutta:
@bmerino95 is this PR still relevant? (it diverged from mainline now and would have to be updated..)

kareninysimba (changes requested):
Hi Brian,
Just a couple of notes:
* Notebooks don't automatically start with the DRAGONS 4.0 kernel.
* Notebook `GMOS_longslit_WhiteDwarf.ipynb` fails when running this cell:
    ```python
    reduce_science = Reduce()
    reduce_science.files.extend(scitarget)
    reduce_science.runr()
    ```
    I get this error:
    ```python
    OSError: No processed flat listed for S20171022S0087_wavelengthSolutionAttached.fits
   ```

Everything else seems to run fine.

**Suggestion for future iterations**:
Have you thought of a way the GHOST DRAGONS notebook might be compressed?
Maybe combining adjacent cells?
Almost 80 code cells seems a little excessive. 
If it was considered before and discarded already, it's fine, nothing critical.

bmerino95:
Hi @kareninysimba, thank you for the feedback! 

- I updated all of the notebooks to open the `DRAGONS-4.0.0 (DL,Py3.12)` kernel by default. 

- Regarding `GMOS_longslit_WhiteDwarf.ipynb`, the error you shared is one that I resolved last year. Can you tell me which Science biases the notebook you ran is downloading? My updated notebook should not include these files: S20171021S0265-269.

Your suggestion to trim the notebooks down is a good one. I think several of the code blocks that generate lists are redundant and can be trimmed. When the next version of DRAGONS is released in the next few months, I can try shortening the notebooks.

kareninysimba (approved):
Hi Brian,
I indeed seemed to have an older version of the GHOST DRAGONS notebook.
I reviewed the new one and in fact there were no issues.
I don't see any other issues!
Merging next...

PR #290, opened 2025-10-03
Update SPARCL+Prospect notebook

weaverba137:
This PR adds a section to the SPARCL+Prospect notebook to show comparisons between SPARCL models and models reconstructed from template coefficients.

In addition a separate issue was discovered where SPARCL records may have been mismatched with metadata from the database due to different ordering. The ordering is now checked for DESI, SDSS, and eBOSS spectra.


weaverba137:
This all looks good. Let's go ahead and merge. If you want to go ahead please do. Otherwise I'll merge in a bit.

PR #293, opened 2025-12-09
Fixed the gmos longslit notebook error

bmerino95:
I realized that the GMOS longslit notebook stopped working and errored out when creating a master flat. I looked into the issue and found it was due to certain scientific biases. When creating a bias with all 10 frames, the science throws an error that it cannot find a proper bias. I checked the Gemini Observation Archive and found that the biases (S0265-S0269) have their QA status set to 'FAIL'.

After some testing, I found that the notebook runs successfully when these files are dropped. The final product looks identical to the previous version, so dropping these files does not affect the final product. 

I ran this updated version of the notebook on GP13 (DRAGONS-4.0.0 (DL, Py3.12)) and verified that it runs without any errors.


bmerino95:
Thanks for pointing that out, @rnikutta! I'll look into this error and let you know when it is ready for testing again.

